PR #4, opened 2016-10-18
RFC use Term type for parsing Formulas

kleinschmidt:
This PR changes how Formulas are parsed. Instead of manually manipulating the `Expr`, this converts it into a tree of `Term`s. These are parametric on the type of node, e.g. `Term{:+}`. These are combined into a reduced tree, adding children one at a time, and using dispatch to implement special rules like the associative and distributive properties, expansion of `*`, etc.

This results in cleaner parsing code, which is also (I think) more extensible, since people need only add additional `add_children!` methods to implement whatever behavior they'd like. It also, I hope, will make for writing more flexible `ModelMatrix` code, since you can dispatch on each of the Terms and the data source type to create the actual columns. But I haven't really dug into that yet.

I'm not 100% satisfied with this design but it passes all the tests and I think is ready for input.  The main thing I'm unhappy with is how leaf nodes ("evaluation terms") are handled. These implemented using a special `:eval` place holder, with the variable name symbol as the only child. Originally I had used the symbol itself as the parameter, but this is a terrible idea because you need to compile special methods to handled every variable (and so performance goes to crap for formulas with many variables).

### DIFF
--- a/src/formula.jl
+++ b/src/formula.jl
@@ -48,9 +48,9 @@ end
 typealias InterceptTerm Union{Term{0}, Term{-1}, Term{1}}
 
 ## equality of Terms
-import Base.==
-=={G,H}(::Term{G}, ::Term{H}) = false
-=={H}(a::Term{H}, b::Term{H}) = a.children == b.children
+Base.:(==){G,H}(::Term{G}, ::Term{H}) = false
+Base.:(==){H}(a::Term{H}, b::Term{H}) = a.children == b.children
+Base.hash{H}(t::Term{H}, h::UInt) = hash(t.children, hash(H, h))
 
 ## display of terms
 function Base.show{H}(io::IO, t::Term{H})
@@ -175,7 +175,7 @@ Base.:(==)(t1::Terms, t2::Terms) = all(getfield(t1, f)==getfield(t2, f) for f in
 function Terms(f::Formula)
     ## start by raising everything on the right-hand side by converting
     rhs = sort!(Term{:+}(Term(f.rhs)))
-    terms = rhs.children
+    terms = filter(isfe, unique(rhs.children))
 
     ## detect intercept
     is_intercept = [isa(t, InterceptTerm) for t in terms]
--- a/src/formula.jl
+++ b/src/formula.jl
@@ -24,9 +24,139 @@ macro ~(lhs, rhs)
     return ex
 end
 
-#
-# TODO: implement contrast types in Formula/Terms
-#
+function Base.show(io::IO, f::Formula)
+    print(io,
+          string("Formula: ",
+                 f.lhs == nothing ? "" : f.lhs, " ~ ", f.rhs))
+end
+
+
+
+
+## Define Terms type that manages formula parsing and extension.
+
+abstract AbstractTerm
+
+type Term{H} <: AbstractTerm
+    children::Vector{Union{Term,Symbol}}
+
+    Term() = new(Term[])
+    Term(children::Vector) = add_children!(new(Term[]), children)
+    Term(child::Symbol) = new([child])
+end
+
+typealias InterceptTerm Union{Term{0}, Term{-1}, Term{1}}
+
+## equality of Terms
+import Base.==
+=={G,H}(::Term{G}, ::Term{H}) = false
+=={H}(a::Term{H}, b::Term{H}) = a.children == b.children
+
+## display of terms
+function Base.show{H}(io::IO, t::Term{H})
+    print(io, string(H))
+    if length(t.children) > 0
+        print(io, "(")
+        print(io, join(map(string, t.children), ", "))
+        print(io, ")")
+    end
+end
+Base.show(io::IO, t::Term{:eval}) = print(io, string(t.children[1]))
+## show ranef term:
+Base.show(io::IO, t::Term{:|}) = print(io, "(", t.children[1], " | ", t.children[2], ")")
+
+## Converting to Term:
+
+## Symbols are converted to :eval terms (leaf nodes)
+Base.convert(::Type{Term}, s::Symbol) = Term{:eval}(s)
+
+## Integers to intercept terms
+function Term(i::Integer)
+    i == 0 || i == -1 || i == 1 || error("Can't construct term from Integer $i")
+    Term{i}()
+end
+
+## no-op constructor
+Term{H}(t::Term{H}) = t
+
+## convert from one head type to another
+(::Type{Term{H}}){H,J}(t::Term{J}) = add_children!(Term{H}(), t, [])
+
+## Expressions are recursively converted to Terms, depth-first, and then added
+## as children.  Specific `add_children!` methods handle special cases like
+## associative and distributive operators.
+function Base.convert(::Type{Term}, ex::Expr)
+    ex.head == :call || error("non-call expression detected: '$(ex.head)'")
+    add_children!(Term{ex.args[1]}(), [Term(a) for a in ex.args[2:end]])
+end
+
+
+## Adding children to a Term
+
+## General strategy: add one at a time to allow for dispatching on special
+## cases, but also keep track of the rest of the children being added because at
+## least the distributive rule requires that context.
+add_children!(t::Term, new_children::Vector) =
+    isempty(new_children) ?
+    t :
+    add_children!(t::Term, new_children[1], new_children[2:end])
+
+function add_children!(t::Term, c::Term, others::Vector)
+    push!(t.children, c)
+    add_children!(t, others)
+end
+
+## special cases:
+## Associative rule
+add_children!(t::Term{:+}, new_child::Term{:+}, others::Vector) =
+    add_children!(t, cat(1, new_child.children, others))
+add_children!(t::Term{:&}, new_child::Term{:&}, others::Vector) =
+    add_children!(t, cat(1, new_child.children, others))
+
+## Distributive property
+## &(a..., +(b...), c...) -> +(&(a..., b_i, c...)_i...)
+add_children!(t::Term{:&}, new_child::Term{:+}, others::Vector) =
+    Term{:+}([add_children!(deepcopy(t), c, others) for c in new_child.children])
+
+## Expansion of a*b -> a + b + a&b
+expand_star(a::Term,b::Term) = Term{:+}([a, b, Term{:&}([a,b])])
+add_children!(t::Term, new_child::Term{:*}, others::Vector) =
+    add_children!(Term{:+}(reduce(expand_star, new_child.children)), others)
+
+## Handle - for intercept term -1, and throw error otherwise
+add_children!(t::Term{:-}, children::Vector) =
+    isa(children[2], Term{1}) ?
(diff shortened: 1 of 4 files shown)


src/formula.jl
```diff
@@ -24,9 +24,138 @@ macro ~(lhs, rhs)
     return ex
 end
 
-#
-# TODO: implement contrast types in Formula/Terms
-#
+function Base.show(io::IO, f::Formula)
+    print(io,
+          string("Formula: ",
+                 f.lhs == nothing ? "" : f.lhs, " ~ ", f.rhs))
```
ararslan: You could put this all on one line and still be under the usual cutoff. Also I think it's typically preferred to use `===` when comparing against `nothing`.

src/formula.jl
```diff
@@ -24,9 +24,138 @@ macro ~(lhs, rhs)
...
+Base.show(io::IO, t::Term{:|}) = print(io, "(", t.children[1], " | ", t.children[2], ")")
+
+## Converting to Term:
+
+## Symbols are converted to :eval terms (leaf nodes)
+Base.convert(::Type{Term}, s::Symbol) = Term{:eval}(s)
+
+## Integers to intercept terms
+function Term(i::Integer)
+    i == 0 || i == -1 || i == 1 || error("Can't construct term from Integer $i")
```
ararslan: `throw`ing an `ArgumentError` might be more appropriate in this situation.

src/formula.jl
```diff
@@ -42,158 +171,42 @@ end
...
-## Otherwise return the expression
-function ex_or_args(ex::Expr,s::Symbol)
-    if ex.head != :call error("Non-call expression encountered") end
-    if ex.args[1] == s
-        ## recurse in case there are more :calls of s below
-        return vcat(map(x -> ex_or_args(x, s), ex.args[2:end])...)
-    else
-        ## not a :call to s, return condensed version of ex
-        return condense(ex)
+    haslhs = f.lhs != nothing
```
ararslan: `!==`

I'm kind of wondering whether we would benefit from having the right- and left-hand sides of the formula be `Nullable`, so these checks are just `isnull`.

src/formula.jl
```diff
@@ -42,158 +171,42 @@ end
...
-    excp.args = vcat(a1, map(x->ex_or_args(x,a1), ex.args[2:end])...)
-    excp
-end
-condense(a::Any) = a
 
-## always return an ARRAY of terms
-getterms(ex::Expr) = (ex.head == :call && ex.args[1] == :+) ? ex.args[2:end] : Expr[ex]
-getterms(a::Any) = Any[a]
+    evalterm_sets = [Set(x) for x in evalterms]
+    evalterms = unique(vcat(evalterms...))
```
ararslan: `reduce(vcat, evalterms)` avoids the performance penalty associated with splatting

src/formula.jl
```diff
@@ -42,158 +171,42 @@ end
...
+    
+    factors = Int8[t in s for t in evalterms, s in evalterm_sets]
+    non_redundants = fill(false, size(factors)) # initialize to false
```
ararslan: `fill(false, n)` -> `falses(n)`?

src/formula.jl
nalimilan: `reduce` will grow the vector repeatedly, which isn't great either. Here you don't actually need to build the full vector just to extract the unique values. I think repeated calls to `union!` should do what you want most efficiently. Or if performance doesn't really matter, `reduce(union, evalterms)`.

src/formula.jl
```diff
@@ -24,9 +24,133 @@ macro ~(lhs, rhs)
...
+Base.convert(::Type{Term}, s::Symbol) = Term{:eval}(s)
+
+## Integers to intercept terms
+function Term(i::Integer)
+    i in -1:1 || throw(ArgumentError("Can't construct term from Integer $i"))
+    Term{i}()
+end
+
+## no-op constructor
+Term{H}(t::Term{H}) = t
```
nalimilan: No need for `H` here.

src/formula.jl
```diff
@@ -24,9 +24,133 @@ macro ~(lhs, rhs)
...
+add_children!(t::Term{:&}, new_child::Term{:&}, others::Vector) =
+    add_children!(t, cat(1, new_child.children, others))
+
+## Distributive property
+## &(a..., +(b...), c...) -> +(&(a..., b_i, c...)_i...)
+add_children!(t::Term{:&}, new_child::Term{:+}, others::Vector) =
+    Term{:+}([add_children!(deepcopy(t), c, others) for c in new_child.children])
+
+## Expansion of a*b -> a + b + a&b
+expand_star(a::Term,b::Term) = Term{:+}([a, b, Term{:&}([a,b])])
```
nalimilan: Missing spaces after commas.

src/formula.jl
```diff
@@ -24,9 +24,133 @@ macro ~(lhs, rhs)
...
+add_children!(t::Term, new_child::Term{:*}, others::Vector) =
+    add_children!(t, cat(1, reduce(expand_star, new_child.children), others))
+
+## Handle - for intercept term -1, and throw error otherwise
+add_children!(t::Term{:-}, children::Vector) =
+    isa(children[2], Term{1}) ?
+    Term{:+}([children[1], Term{-1}()]) :
+    error("invalid subtraction of $(children[2]); subtraction only supported for -1")
+
+## sorting term by the degree of its children: order is 1 for everything except
```
nalimilan: "sorting a term's children by their degree" would be more accurate.

src/formula.jl
```diff
@@ -24,9 +24,133 @@ macro ~(lhs, rhs)
...
+degree(::Term) = 1
+degree(::InterceptTerm) = 0
+
+function Base.sort!(t::Term)
+    sort!(t.children, by=degree)
+    return t
+end
+
+################################################################################
+## This duplicates the functionality of the DataFrames.Terms type:
```
nalimilan: The DataFrames type is supposed to be removed, isn't it? If so, no need to mention it in the code.

test/term.jl
```diff
@@ -0,0 +1,31 @@
...
+
+## Expand * to main effects + interactions
+@test Term{:+}(Term(:(a*b))) == Term(:(a+b+a&b))
+@test sort!(Term{:+}(Term(:(a*b*c)))) == Term(:(a+b+c+a&b+a&c+b&c+a&b&c))
+@test Term(:(a + b*c)) == Term(:(a + b + c + b&c))
+@test Term(:(a*b + c)) == Term(:(a + b + a&b + c))
+
+## printing Terms:
+@test string(Term(:a)) == "a"
+@test string(Term(:(a+b))) == "+(a, b)"
```
nalimilan: Wouldn't it be better to use infix operators? Or is that too much work?

src/formula.jl
```diff
@@ -42,158 +166,42 @@ end
...
-end
-condense(a::Any) = a
 
-## always return an ARRAY of terms
-getterms(ex::Expr) = (ex.head == :call && ex.args[1] == :+) ? ex.args[2:end] : Expr[ex]
-getterms(a::Any) = Any[a]
+    evalterm_sets = [Set(x) for x in evalterms]
+    evalterms = unique(reduce(vcat, [], evalterms))
+    
+    factors = Int8[t in s for t in evalterms, s in evalterm_sets]
```
nalimilan: Why not `Bool`?

src/formula.jl
```diff
@@ -42,158 +166,42 @@ end
...
+    non_redundants = falses(size(factors)) # initialize to false
```
nalimilan: `falses` creates a `BitArray`, which is going to be converted by the constructor. So `fill(falses, size(factors))` is actually better.

src/formula.jl
```diff
@@ -42,158 +166,42 @@ end
...
 
-ord(ex::Expr) = (ex.head == :call && ex.args[1] == :&) ? length(ex.args)-1 : 1
-ord(a::Any) = 1
+    Terms(terms, evalterms, factors, non_redundants, degrees, haslhs, hasintercept)
```
nalimilan: Useless line break.

src/formula.jl
```diff
@@ -24,9 +24,133 @@ macro ~(lhs, rhs)
...
 end
 
-#
-# TODO: implement contrast types in Formula/Terms
-#
+Base.show(io::IO, f::Formula) =
+    print(io, string("Formula: ", f.lhs === nothing ? "" : f.lhs, " ~ ", f.rhs))
+
+
+## Define Terms type that manages formula parsing and extension.
```
nalimilan: `Term` type (no plural).

src/formula.jl
```diff
@@ -24,9 +24,133 @@ macro ~(lhs, rhs)
...
+    children::Vector{Union{Term,Symbol}}
+
+    Term() = new(Term[])
+    Term(children::Vector) = add_children!(new(Term[]), children)
+    Term(child::Symbol) = new([child])
+end
+
+typealias InterceptTerm Union{Term{0}, Term{-1}, Term{1}}
+
+## equality of Terms
```
nalimilan: Same here.

src/formula.jl
ararslan: Oh interesting, I didn't realize that. My bad for suggesting `falses`.

nalimilan: Yeah, it's a bit confusing and stands out in the API. I'd rather do this via `falses(BitArray, ...)`.

test/term.jl
kleinschmidt: I think it would be better, too. Marginally more work but probably worth it, this is just what I hacked together initially...

src/formula.jl
kleinschmidt: Holdover from the DataFrames.jl version...

src/formula.jl
kleinschmidt: This was just a note to myself that this is the "old bit". I think, ultimately, that we can replace the `Terms` type and work with the tree of `Term` directly when constructing model matrices.

test/term.jl
nalimilan: OK, this can wait for another PR.

ararslan:
Bump!

kleinschmidt:
I've thought a bit more about this in the context of generating model matrices. My long-term vision for this has roughly been that a `Term` is the atomic unit that describes how column(s) are generated from data, or from other `Term`s.  For instance, a `Term{:+}` says, generate columns for each of my children and then horizontally concatenate them.  A `Term` tree could then replace the concept of the old `Terms` type, and possibly also `ModelFrame` (transforming leaf nodes to point to the corresponding data source, although it's less clear to me that this is a good idea).  We'd handle things like encoding contrasts and checking redundancy by actually transforming the underlying `Term`s themselves.  This will obviously take some doing to even see whether it works.

It would still be possible to use an abstraction like this without _immediately_ parsing expressions into Terms, but I don't really see the advantage of that (other than possibly performance), especially since there's already a fair amount of transformations that happen in that parsing process (e.g., applying distributive property, expanding `*`).

kleinschmidt:
This is superseded by #54 and #71

PR #6, opened 2016-11-18
Document contrasts and formulas etc.

kleinschmidt:
Here's a first pass at documentation for Formulas, ModelFrame/Matrix, and Contrasts.


docs/src/formula.md
```diff
@@ -0,0 +1,76 @@
...
+DocTestSetup = quote
+    using StatsModels
+end
+```
+
+# Modeling tabular data
+
+Most statistical models require that data be represented as a Matrix-like
+collection of a single numeric type.  Much of the data we want to model,
+however, is __tabular data__, where data is represented as a collection of
```
ararslan: Being ultra pedantic here, but "data" is actually plural. So that should be "where data are" rather than "where data is."

docs/src/formula.md
```diff
@@ -0,0 +1,76 @@
...
+absence of a constant intercept term, respectively---and variables like `x`,
+which will evaluate to the data source column with that name as a symbol (`:x`).
+
+Individual variables can be combined into _interaction terms_ with `&`, as in
+`x&z`, which will evaluate to the product of the columns named `:x` and `:z`.
+Because it's often convenient to include main effects and interactions for a
+number of variables, the `*` operator will expand in this way
+
+```jldoctest
+# Using the internal StatsModels.Terms to trigger expansion of the Formula...
```
ararslan: Do `jldoctest`s allow comments like this? (cc @MichaelHatherly)

src/contrasts.jl
```diff
@@ -22,7 +22,7 @@ instantiation optionally includes the levels to generate columns for and the bas
 level. If not specified these will be taken from the data when a `ContrastsMatrix` is
 generated (during `ModelFrame` construction).
 
-# Constructors
+**Constructors**
```
ararslan: It seems more natural IMO to separate sections like this by using markdown titles (doesn't have to be h1) rather than bold. Typically it's `#` for main title, `##` for sections, and `###` for subsections.

src/contrasts.jl
```diff
@@ -83,12 +90,37 @@ type ContrastsMatrix{C <: AbstractContrasts, T}
 end
 
 """
-    ContrastsMatrix{C<:AbstractContrasts}(contrasts::C, levels::AbstractVector)
+An instantiation of a contrast coding strategy for particular levels
 
-Compute contrasts matrix for given data levels.
+This type is used internally for generating model matrices based on categorical
+data, and __most users will not need to deal with it directly__.  It's exported
+for developers that need to extend these facilities.  Conceptually, a
```
ararslan: Does it have to be exported in order to be extended by another package? If not, we should probably just not export it since it isn't intended to be user-facing.

ararslan:
Overall 💯, excellent work!

docs/src/formula.md
MichaelHatherly: They do, but only really by accident. Rather do the following if full line comments are needed

```
julia> # Using the internal StatsModels.Terms to trigger expansion of the Formula...
       Formula(StatsModels.Terms(y ~ 1 + x*y))
Formula: y ~ 1 + x + y + x & y
```

since in the REPL you can't actually write something like

```
# A comment...
julia> ...
```

docs/src/formula.md
```diff
@@ -0,0 +1,76 @@
...
+Formula: y ~ 1 + x
+```
+
+The left side of a formula conventionally represents _dependent_ variables, and
+the right side _independent_ variables (or regressors).  _Terms_ are separated
+by `+`.  Basic terms are the integers `1` or `0`---evaluated as the presence or
+absence of a constant intercept term, respectively---and variables like `x`,
+which will evaluate to the data source column with that name as a symbol (`:x`).
+
+Individual variables can be combined into _interaction terms_ with `&`, as in
```
MichaelHatherly: Underscores for bold and italic aren't actually supported by our markdown parser &mdash; should use `*` instead.

.travis.yml
```diff
@@ -13,6 +13,9 @@ script:
   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
   - julia -e 'Pkg.clone(pwd()); Pkg.checkout("DataFrames", "dfk/statsmodel-purge"); Pkg.build("StatsModels"); Pkg.test("StatsModels"; coverage=true)'
 after_success:
+  # build and deploy documentation with Documenter.jl
+  - openssl aes-256-cbc -K $encrypted_a9d536040c76_key -iv $encrypted_a9d536040c76_iv -in docs/.documenter.enc -out docs/.documenter -d
```
MichaelHatherly: The `.documenter.enc` method for deploying docs from travis has been deprecated in favour of a much simpler approach, which is documented here: https://juliadocs.github.io/Documenter.jl/latest/man/hosting.html#SSH-Deploy-Keys-1

.travis.yml
```diff
@@ -13,6 +13,9 @@ script:
...
+  - julia -e 'cd(Pkg.dir("StatsModels")); Pkg.clone("https://github.com/MichaelHatherly/Documenter.jl"); include(joinpath("docs", "make.jl"))'
```
MichaelHatherly: Rather do a `Pkg.add("Documenter")` here instead of cloning from my personal fork of the package since it'll be a bit unstable and buggy more than likely.

docs/make.jl
```diff
@@ -0,0 +1,8 @@
+using Documenter, StatsModels
+
+makedocs()
```
MichaelHatherly: Might be worth considering the new HTML format supported by Documenter, https://juliadocs.github.io/Documenter.jl/latest/man/guide.html#Output-formats-1, since it's already being used by other JuliaStats packages (`DataFrames`) and might be nice to have a consistent documentation style across the entire organisation.

docs/src/contrasts.md
```diff
@@ -0,0 +1,77 @@
...
+Each contrast coding strategy maps a categorical vector with $k$ levels onto
+$k-1$ linearly independent model matrix columns.
+
+## How to specify contrast coding
+
+The default contrast coding strategy is `DummyCoding`.  To override this, use
+the `contrasts` argument when constructing a `ModelFrame`:
+
+```julia
+mf = ModelFrame(y ~ 1 + x, df, contrasts = Dict(:x => EffectsCoding()))
```
nalimilan: `EffectsCoding` hasn't been mentioned yet. Maybe better list first the supported strategies, and only then say how to use them?

docs/src/contrasts.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+## Further details
+
+### Categorical variables in `Formula`s
+
+Generating model matrices from multiple variables, some of which are
+categorical, requires special care.  The reason for this is that rank-$k-1$
+contrasts are appropriate for a categorical variable with $k$ levels when it is
+_redundant_ with lower-order terms: using rank-$k$ will often result in a
```
nalimilan: Shouldn't "often" be "always" (or nothing)?

docs/src/contrasts.md
```diff
@@ -0,0 +1,77 @@
...
+rank-deficient model matrix which leads to a model that can't be identified.  A
+categorical variable in a term is _redundant_ when the term obtained by dropping
```
nalimilan: I don't really understand this sentence. It doesn't seem to work with the first example below, or I'm missing something.

docs/src/contrasts.md
```diff
@@ -0,0 +1,77 @@
...
+that variable is identical to a term already present in the
+formula.[^implicit-terms]
+
+For example: 
+* In `y ~ 1 + x`, `x` is redundant with the intercept `1`.
```
nalimilan: "partly redundant"?

Also repeat that `x` is categorical.

docs/src/contrasts.md
nalimilan: Reading the examples carefully, I see what you mean. I think the term "redundant" is what confused me: I would interpret it as meaning that a term doesn't add anything to the model; these terms definitely add something. Could we find a better name, like "repeated"?

docs/src/formula.md
```diff
@@ -0,0 +1,76 @@
...
+`x&z`, which will evaluate to the product of the columns named `:x` and `:z`.
+Because it's often convenient to include main effects and interactions for a
+number of variables, the `*` operator will expand in this way
```
nalimilan: Give an example?

nalimilan: Scratch that, just ":" would be enough.

docs/src/formula.md
```diff
@@ -0,0 +1,76 @@
...
+
+```jldoctest
+# Using the internal StatsModels.Terms to trigger expansion of the Formula...
+julia> Formula(StatsModels.Terms(y ~ 1 + x*y))
+Formula: y ~ 1 + x + y + x & y
+```
+
+This applies to higher-order interactions, too: `x*y*z` expands to the main
+effects, all two-way interactions, and the three way interaction `x&y&z`.
```
nalimilan: Show the full expansion for clarity.

src/contrasts.jl
```diff
@@ -33,31 +33,38 @@ C(base = ::Any)                         # specify base level
...
-`ContrastsMatrix`. Any mismatch will result in an error, because missing data
-levels would lead to empty columns in the model matrix, and missing contrast
-levels would lead to empty or undefined rows.
+**Arguments**
 
-You can also specify the base level of the contrasts. The actual interpretation
-of this depends on the particular contrast type, but in general it can be
-thought of as a "reference" level.  It defaults to the first level.
+* `levels`: Optionally, the data levels can be specified here.  This allows you
+  to specify the order of the levels.  If specified, the levels will be checked
```
nalimilan: "the levels provided here will be checked against the levels actually present in the data"?

src/contrasts.jl
```diff
@@ -33,31 +33,38 @@ C(base = ::Any)                         # specify base level
...
+  against the levels when the `ContrastsMatrix` is constructed. Any mismatch
+  will result in an error, because missing data levels would lead to empty
```
nalimilan: "levels missing from the data" (to avoid the confusion with "missing data")

src/contrasts.jl
nalimilan: No, it doesn't. So indeed we can keep this unexported.

src/contrasts.jl
```diff
@@ -196,11 +228,22 @@ end
 Coding that generates one indicator (1 or 0) column for each level,
 __including__ the base level.
 
-Needed internally when a term is non-redundant with lower-order terms (e.g., in
-`~0+x` vs. `~1+x`, or in the interactions terms in `~1+x+x&y` vs. `~1+x+y+x&y`. In the
-non-redundant cases, we can expand x into `length(levels(x))` columns
-without creating a non-identifiable model matrix (unless the user has done
-something foolish in specifying the model, which we can't do much about anyway).
+Not exported but included here for the sake of completeness.
```
nalimilan: If we don't export the other functions either, I'd remove this line.

src/modelframe.jl
```diff
@@ -1,3 +1,50 @@
+"""
+Wrapper which combines Formula (Terms) and an AbstractDataFrame
+
+This wrapper encapsulates all the information that's required to transform data
+of the same structure as the wrapped `DataFrame` into a model matrix.  This goes
```
nalimilan: "data frame" since other types are potentially supported.

src/modelframe.jl
```diff
@@ -1,3 +1,50 @@
...
+variables are categorical and determines the appropriate contrasts to use, and
+then creates the necessary contrasts matrices and stores the results.
+
+**Constructors**
+
+```julia
+ModelFrame(f::Formula, df::AbstractDataFrame; contrasts::Dict = Dict())
+ModelFrame(ex::Expr, d::AbstractDataFrame; contrasts::Dict = Dict())
+ModelFrame(terms::Terms, df::AbstractDataFrame; contrasts::Dict = Dict())
+# Internal constructors:
```
nalimilan: Inner?

src/modelframe.jl
```diff
@@ -1,3 +1,50 @@
...
+ModelFrame(df::AbstractDataFrame, terms::Terms, missing::BitArray)
+ModelFrame(df::AbstractDataFrame, terms::Terms, missing::BitArray, contrasts::Dict{Symbol, ContrastsMatrix})
+```
+
+**Arguments**
+
+* `f`: Formula whose left hand side is the _response_ and right hand side are
```
nalimilan: Also maybe repeat the argument types here and below, this often clarifies a lot what it's supposed to contain.

src/modelframe.jl
```diff
@@ -1,3 +1,50 @@
...
+  the predictors.
+* `df`: The data being modeled.  This is used at this stage to determine which
+  variables are categorical, and otherwise held for [`ModelMatrix`](@ref).
+* `contrasts`: An optional Dict of contrast codings for each categorical
+  variable.  Any unspecified variables will have [`DummyCoding`](@ref).  As a
+  keyword argument, these can be either instances of a subtype of
+  [`AbstractContrasts`](@ref), or a [`ContrastsMatrix`](@ref).  For the internal
```
nalimilan: Inner here too?

nalimilan: (And below)

src/modelframe.jl
```diff
@@ -1,3 +1,50 @@
...
+  constructor, they must be [`ContrastsMatrix`](@ref)es.
+* `ex`: An expression will be converted into a `Formula`.
```
nalimilan: which

src/modelframe.jl
```diff
@@ -1,3 +1,50 @@
...
+* `terms`: For internal constructor, the parsed `Terms` from the `Formula`.
+* `missing`: For internal constructor, a `BitArray` of which rows of the data
```
nalimilan: "a `BitArray` indicating whether each row of `df` contains missing values"?

src/modelframe.jl
```diff
@@ -79,6 +117,15 @@ function evalcontrasts(df::AbstractDataFrame, contrasts::Dict = Dict())
     return evaledContrasts
 end
 
+## Default NULL handler.  Others can be added as keyword arguments
```
nalimilan: Unrelated?

src/modelmatrix.jl
```diff
@@ -1,6 +1,22 @@
 
 typealias AbstractFloatMatrix{T<:AbstractFloat} AbstractMatrix{T}
 
+"""
+Convert a `ModelFrame` into a numeric matrix suitable for modeling
```
nalimilan: If you document the constructors, give the signatures first like any other function? I'm not sure what are the conventions for this.

.travis.yml
kleinschmidt: Hm, those are the instructions I followed. Or, rather, I ran `Travis.genkeys()` and did what it told me to do. I'm running Documenter.jl 0.7.2.

kleinschmidt: Also, in case it's important, I have v1.8.4 (the latest release) of the travis cli, installed via

    $ gem install travis -v 1.8.4 --no-rdoc --no-ri

(as per the [instructions on their github page](https://github.com/travis-ci/travis.rb#installation))

kleinschmidt: My bad, apparently I hadn't actually updated `Documenter`. Things work as intended now.

src/contrasts.jl
kleinschmidt: Good points. Turns out I hadn't even exported it ¯\_(ツ)_/¯

docs/make.jl
kleinschmidt: Ah I see that now.  I was going by the last release apparently.

kleinschmidt: That material theme is real pretty though.

src/modelmatrix.jl
kleinschmidt: I was working off DataFrames.jl, which gives a short summary, then lists constructors and their arguments. I'm not particularly wedded to this way but I'm also not aware of any other standard besides that

src/contrasts.jl
kleinschmidt: We do export them. They're the user-facing interface for specifying contrast coding strategy.

docs/src/contrasts.md
nalimilan: No, definitely not "dangerous". :-)

Maybe "partly redundant"? Or maybe better,I think R says "aliased" in this case. So maybe "aliased (i.e. partly redundant) with" the first time you talk about this.

src/modelmatrix.jl
nalimilan: OK, fine then. We need to improve the manual about this to choose a convention.

src/contrasts.jl
```diff
@@ -224,6 +266,17 @@ regression model, dummy coding leads to an intercept that is the mean of the
 dependent variable for base level.
 
 Also known as "treatment coding" (`contr.treatment` in R) or "one-hot encoding".
```
nalimilan: While you're at it, I think we should stop mentioning R, as we don't want to list every possible software. Giving the terminology they use is enough. Also, I would mention "Dummy coding" in the first (summary) line: even if it's in the type name, I would find it clearer. Same remarks for other contrasts.

docs/src/formula.md
```diff
@@ -40,8 +40,7 @@ Because it's often convenient to include main effects and interactions for a
 number of variables, the `*` operator will expand in this way
 
 ```jldoctest
-# Using the internal StatsModels.Terms to trigger expansion of the Formula...
-julia> Formula(StatsModels.Terms(y ~ 1 + x*y))
+julia> Formula(StatsModels.Terms(y ~ 1 + x*y)) # Parse by converting to Terms
```
nalimilan: The old comment was more explicit. Maybe keep it, but moving it outside of the code block?

docs/src/contrasts.md
```diff
@@ -9,6 +9,13 @@ modeling, `StatsModels` implements a variety of **contrast coding strategies**.
 Each contrast coding strategy maps a categorical vector with $k$ levels onto
 $k-1$ linearly independent model matrix columns.
 
+The following strategies are implemented:
```
nalimilan: Thinking about it again, "strategy" sounds a bit unusual in this domain. "Coding scheme" or "coding system" may be more common (e.g. http://www.ats.ucla.edu/stat/r/library/contrast_coding.htm).

src/modelframe.jl
```diff
@@ -16,26 +16,27 @@ then creates the necessary contrasts matrices and stores the results.
...
-  the predictors.
-* `df`: The data being modeled.  This is used at this stage to determine which
-  variables are categorical, and otherwise held for [`ModelMatrix`](@ref).
-* `contrasts`: An optional Dict of contrast codings for each categorical
+* `f::Formula`: Formula whose left hand side is the *response* and right hand
+  side are the *predictors*.
+* `df::AbstractDataFrame`: The data being modeled.  This is used at this stage
+  to determine which variables are categorical, and otherwise held for
+  [`ModelMatrix`](@ref).
+* `contrasts::Dict`: An optional Dict of contrast codings for each categorical
```
nalimilan: "contrast coding schemes" would be better (IIRC sometimes "contrast coding" refers to non-standard/complex contrast schemes).

src/contrasts.jl
```diff
@@ -233,7 +233,7 @@ levels needs to be converted into ``k`` model matrix columns instead of the
 standard ``k-1``.  This occurs when there are missing lower-order terms, as in
 discussed below in [Categorical variables in Formulas](@ref).
 
-**Examples**
+# Examples
 
 ```julia
```
nalimilan: I think using blocks is deprecated: just use a four-space indent for code. (DataFrames needs updating to the new conventions.)

nalimilan:
Looks good, but since I'm greedy I've commented about further possible improvements.

src/contrasts.jl
kleinschmidt: Both the Documenter.jl style guide and the julia manual section on documentation suggest using blocks.

src/contrasts.jl
kleinschmidt: I'm not sure about dropping mention of R. On the one hand, I'm sympathetic to the fact that we can't realistically mention every single alternative. On the other hand, I suspect that many users will be coming from R, but that might be biased by my own experience :)

docs/src/contrasts.md
kleinschmidt: But "dangerous" sounds so fun!

I actually like something like "the variable aliases the remainder of the term". Then we can replace "redundant variable" with "aliasing variable" and "non-redundant" with "non-aliasing". How does that sound?

src/contrasts.jl
nalimilan: One issue with always comparing with R is that people will expect things to work the same way, which is not always the case.

src/contrasts.jl
nalimilan: Sorry, I didn't intend to comment on that line. Examples should indeed be blocks (and even better `jldoctest` blocks). I wanted to refer to the constructor signatures on a different line.

src/modelframe.jl
kleinschmidt: oops, yes, I'd made a commit on master that I didn't push.  I'll rebase this all appropriately before merging.

kleinschmidt:
Any final comments here?

.gitignore
```diff
@@ -1,3 +1,5 @@
 *.jl.cov
 *.jl.*.cov
 *.jl.mem
+build
```
MichaelHatherly: This should be `docs/build` and `docs/site` can be removed (since it's not being built with `mkdocs` any more).

docs/mkdocs.yml
```diff
@@ -0,0 +1,31 @@
+site_name: StatsModels.jl
```
MichaelHatherly: Can be removed since not building with `mkdocs`.

src/contrasts.jl
kleinschmidt: That's a good point, and now that I've done a bit more reading I'm not sure if the way that R implements these really does what I thought it does (in terms of interpretation of the coefficients). So I'll take out the explicit mention of R now since our implementation might well change in the future. We might want to consider a "getting started" guide that translates things for R, SAS, etc. users, but I'm not sure that this is the best place for that.

kleinschmidt: That being said, there's one place where people might get confused which we discuss, which is that the base level for effects/sum/simple coding is different in our implementation than in R and SPSS. I think it's a good idea to include that at least.

docs/src/formula.md
kleinschmidt: This could get real ugly 😜, but here's why I think it's appropriate to use "data is" here:

To me (and AFAIK, to many if not most English speakers), "data" is a mass noun, like "water" or "information". The ever-objective wikipedia has a [good summary](https://en.wikipedia.org/wiki/Data_(word)) of where this usage is dominant and where it's not (mostly technical writing, especially in the UK, except computer science and related disciplines). I'd like as much as possible to keep documentation in a colloquial style to make it as approachable as possible, and "data is" is more consistent with that style.

For a little linguistic pedantry, as you know "data" is a plural of Latin "datum". But we don't speak Latin, we speak English (and the Latinate singular form of "datum" is so rare that it hardly registers on Google ngrams). In general, it's really common for grammatical things to get mixed up in borrowings like this. In particular, there are plenty of other words that were plural in Latin that are treated as singular when borrowed in English, like "agenda" (plural of "agendum").

MichaelHatherly:
Also needs a `gh-pages` branch since Documenter won't create one itself. Then just restart the travis build and it should deploy the docs correctly.

kleinschmidt:
Haha thanks, I just finished doing that...missed it the first time I read the docs...

src/contrasts.jl
nalimilan: You mean because the *first* level is skipped in effects coding? Indeed, that's kind of unusual; are there examples of other software doing this?

BTW, this would be nicely solved by printing the (pseudo-)coefficient for the skipped level, which users are likely to compute manually anyway. Do you have opinions on point 2 of https://github.com/JuliaStats/GLM.jl/issues/29?

kleinschmidt: I don't know of any other software packages that do this, but I got sick of explaining to beginners why the coefficient changes sign (and label) when you use `contr.sum` instead of the default `contr.treatment` in R. This way makes the coefficients of effects coded predictors parallel the dummy coded versions: coefficients are generated for levels 1:k-1 which represent how much they deviate from some reference (the grand mean or the first level).

My feeling about point 2 there is that it could be very confusing to print a pseudo coefficient whose interpretation is qualitatively different from the other coefficients. I think it would be a good idea to provide some way to generate pseudo coefficients (e.g., full dummy coefficients) from the output (perhaps as an argument to coefs/coeftable). But I haven't thought carefully about how easy this is to do when there are multiple predictors.

nalimilan: > My feeling about point 2 there is that it could be very confusing to print a pseudo coefficient whose interpretation is qualitatively different from the other coefficients. I think it would be a good idea to provide some way to generate pseudo coefficients (e.g., full dummy coefficients) from the output (perhaps as an argument to coefs/coeftable). But I haven't thought carefully about how easy this is to do when there are multiple predictors.

AFAIK the interpretation of the omitted coefficient isn't really different from that of the actual coefficients when using effects coding: all of them measure the deviation from the mean associated with a level. OTC, viewing the base level as different from the others is weird, since one would get the same results after choosing a different base. And even for dummy coding, printing the name of the reference together with a `0` value is very useful (you can't interpret the coefficients without this information), especially since in some cases the reference can be `1` (e.g. when using odds ratios).

As regards the actual computation, I'm not sure the presence of multiple predictors would be an issue. Doesn't the omitted coefficient only depend on the contrast system? The tricky part is already handled when you choose whether to use full dummy coding or standard dummy coding.

kleinschmidt: I see, I misunderstood what you meant by pseudo-coefficient. I thought you meant something like the full-rank dummy coefficient, not the analogous coefficient given the coding scheme. That might make sense for something like dummy or effects coding, but what would it be for helmert, or forward/reverse difference coding (which we don't currently implement but might as well)?

In general, I'm a little wary of reporting anything other than the actual fitted values _by default_. I think, however, that we should definitely provide a way of "re-coding" fitted models on the fly (since as long as the contrasts are well-formed than any system is linearly equivalent to any other). But it seems to me that that's something users should have to ask for.

nalimilan: I don't think there's an additional coefficient coefficient to report for forward/reverse difference coding; anyway, coefficient names should include all original levels (as part of the representation of differences). For Helmert coding, we cannot compare the last level with subsequent levels, so there's nothing to compute either; though it could be useful to have just the name of the last level with a `NaN` coefficient, so that you know the name of the last level. I really can't say, as I've only used dummy and effects coding in practice.

We don't need to print this by default. But it would be a very useful information to generate publication-level regression tables, where you always need to give the full information about all levels. These additional coefficients could be printed in a different way from others: for example, the base level for dummy coding could be printed with `0` next to the variable name, after "Reference:".

PR #9, opened 2016-12-15
WIP: Replace infix ~ for formulas with a model macro

### LINKED ISSUE #3: An alternative syntax for formulas?
### What we have

Currently formulas are constructed using `@~`, which as far as I know is Julia's only infix macro. This allows us to mirror R's formulas, which take advantage of R's nonstandard evaluation, and avoid prefixing variable names with `:`.

However, this is a special case that Julia makes in parsing that it—at least in my opinion—shouldn't have to do to allow Julia to look like R in one scenario. It would be nice if Julia could avoid having to parse _anything_ as an infix macro. Plus, there surely is a more Julian approach!
### A solution

A while back, @dmbates suggested on the mailing list that we replace our current formula syntax with `Pair`s. Based on my understanding, this would look like `:y => :(1 + x1 + x2)`. @StefanKarpinski alternatively suggested the use of a `@model` macro, which would take an expression and construct a formula or model object, which could then be passed to `fit` or what have you. I imagine this looking something like

``` julia
fit(LinearRegressionModel, @model y ~ 1 + x1 + x2, ...)
```

This approach allows any separator for the dependent and independent variables, not just `~`, since it simply becomes the `head` in the resulting `Expr`. We could even use `=` for a more SAS-like aesthetic, or `=>` as in `Pair`s.
### A call to action

As we're transitioning the modeling functionality out of DataFrames.jl and into its own package (this one!), we have the opportunity to revisit some design decisions such as this. What do you all think? Should we leave things as-is or blaze a new trail?

ararslan:
Fixes #3 

I've gone with `=>` here since we need something that parses as an infix operator. Once `~` stops being parsed as an infix macro, it's not clear to me whether it will still be parsed as an infix operator at all. Thus I've opted for something we can guarantee is infix.

### DIFF
--- a/src/formula.jl
+++ b/src/formula.jl
@@ -3,9 +3,9 @@
 # Douglas M. Bates, and other contributors.
 
 ## Formulas are written as expressions and parsed by the Julia parser.
-## For example :(y => a + b + log(c))
+## For example :(y ~ a + b + log(c))
 ## In Julia the & operator is used for an interaction.  What would be written
-## in R as y ~ a + b + a:b is written :(y => a + b + a&b) in Julia.
+## in R as y ~ a + b + a:b is written :(y ~ a + b + a&b) in Julia.
 ## The equivalent R expression, y ~ a*b, is the same in Julia
 
 ## The lhs of a one-sided formula is 'nothing'
@@ -49,7 +49,7 @@ end
 Base.:(==)(t1::Terms, t2::Terms) = all(getfield(t1, f)==getfield(t2, f) for f in fieldnames(t1))
 
 function Base.show(io::IO, f::Formula)
-    print(io, "Formula: ", f.lhs === nothing ? "" : f.lhs, " => ", f.rhs)
+    print(io, "Formula: ", f.lhs === nothing ? "" : f.lhs, " ~ ", f.rhs)
 end
 
 # special operators in formulas
--- a/src/formula.jl
+++ b/src/formula.jl
@@ -1,27 +1,27 @@
 # Formulas for representing and working with linear-model-type expressions
-# Original by Harlan D. Harris.  Later modifications by John Myles White
-# and Douglas M. Bates.
+# Original by Harlan D. Harris.  Later modifications by John Myles White,
+# Douglas M. Bates, and other contributors.
 
 ## Formulas are written as expressions and parsed by the Julia parser.
-## For example :(y ~ a + b + log(c))
+## For example :(y => a + b + log(c))
 ## In Julia the & operator is used for an interaction.  What would be written
-## in R as y ~ a + b + a:b is written :(y ~ a + b + a&b) in Julia.
+## in R as y ~ a + b + a:b is written :(y => a + b + a&b) in Julia.
 ## The equivalent R expression, y ~ a*b, is the same in Julia
 
 ## The lhs of a one-sided formula is 'nothing'
 ## The rhs of a formula can be 1
 
 type Formula
-    lhs::@compat(Union{Symbol, Expr, Void})
-    rhs::@compat(Union{Symbol, Expr, Integer})
+    lhs::Union{Symbol, Expr, Void}
+    rhs::Union{Symbol, Expr, Integer}
 end
 
-macro ~(lhs, rhs)
-    ex = Expr(:call,
-              :Formula,
-              Base.Meta.quot(lhs),
-              Base.Meta.quot(rhs))
-    return ex
+macro model(formula)
+    formula.head === :(=>) || error("expected formula separator =>, got $(formula.head)")
+    length(formula.args) == 2 || error("malformed expression in formula")
+    lhs = Base.Meta.quot(formula.args[1])
+    rhs = Base.Meta.quot(formula.args[2])
+    return Expr(:call, :Formula, lhs, rhs)
 end
 
 """
@@ -46,9 +46,7 @@ end
 Base.:(==)(t1::Terms, t2::Terms) = all(getfield(t1, f)==getfield(t2, f) for f in fieldnames(t1))
 
 function Base.show(io::IO, f::Formula)
-    print(io,
-          string("Formula: ",
-                 f.lhs == nothing ? "" : f.lhs, " ~ ", f.rhs))
+    print(io, "Formula: ", f.lhs === nothing ? "" : f.lhs, " => ", f.rhs)
 end
 
 # special operators in formulas
--- a/src/formula.jl
+++ b/src/formula.jl
@@ -16,11 +16,18 @@ type Formula
     rhs::Union{Symbol, Expr, Integer}
 end
 
-macro model(formula)
-    formula.head === :(=>) || error("expected formula separator =>, got $(formula.head)")
-    length(formula.args) == 2 || error("malformed expression in formula")
-    lhs = Base.Meta.quot(formula.args[1])
-    rhs = Base.Meta.quot(formula.args[2])
+macro formula(ex)
+    if ex.head === :macrocall && first(ex.args) === Symbol("@~")
+        length(ex.args) == 3 || error("malformed expression in formula")
+        lhs = Base.Meta.quot(ex.args[2])
+        rhs = Base.Meta.quot(ex.args[3])
+    elseif ex.head === :(~)
+        length(ex.args) == 2 || error("malformed expression in formula")
+        lhs = Base.Meta.quot(ex.args[1])
+        rhs = Base.Meta.quot(ex.args[2])
+    else
+        error("expected formula separator ~, got $(ex.head)")
+    end
     return Expr(:call, :Formula, lhs, rhs)
 end
 
--- a/src/formula.jl
+++ b/src/formula.jl
@@ -17,14 +17,10 @@ type Formula
 end
 
 macro formula(ex)
-    if ex.head === :macrocall && first(ex.args) === Symbol("@~")
+    if (ex.head === :macrocall && ex.args[1] === Symbol("@~")) || (ex.head === :call && ex.args[1] === :(~))
         length(ex.args) == 3 || error("malformed expression in formula")
         lhs = Base.Meta.quot(ex.args[2])
         rhs = Base.Meta.quot(ex.args[3])
-    elseif ex.head === :(~)
-        length(ex.args) == 2 || error("malformed expression in formula")
-        lhs = Base.Meta.quot(ex.args[1])
-        rhs = Base.Meta.quot(ex.args[2])
     else
         error("expected formula separator ~, got $(ex.head)")
     end
--- a/docs/src/contrasts.md
+++ b/docs/src/contrasts.md
@@ -22,7 +22,7 @@ The default contrast coding system is `DummyCoding`.  To override this, use
 the `contrasts` argument when constructing a `ModelFrame`:
 
 ```julia
-mf = ModelFrame(y ~ 1 + x, df, contrasts = Dict(:x => EffectsCoding()))
+mf = ModelFrame(@formula(y ~ 1 + x), df, contrasts = Dict(:x => EffectsCoding()))
 ```
 
 To change the contrast coding for one or more variables in place, use
--- a/docs/src/formula.md
+++ b/docs/src/formula.md
@@ -72,6 +72,13 @@ julia> Formula(StatsModels.Terms(@formula(y ~ 1 + (a+b) * c)))
 Formula: y ~ 1 + a + b + c + c & a + c & b
 ```
 
+You may be wondering why formulas in Julia require a macro, while in R they appear "bare."
+R supports nonstandard evaluation, allowing the formula to remain an unevaluated object
+while its terms are parsed out. Julia uses a much more standard evaluation mechanism,
+making this impossible using normal expressions. However, unlike R, Julia provides macros to
+explicitly indicate when code itself will be manipulated before it's evaluated. By constructing
+a formula using a macro, we're able to provide convenient, R-like syntax and semantics.
+
 ## The `ModelFrame` and `ModelMatrix` types
 
 The main use of `Formula`s is for fitting statistical models based on tabular
--- a/docs/src/formula.md
+++ b/docs/src/formula.md
@@ -21,12 +21,15 @@ goal is to support any tabular data format that adheres to a minimal API,
 ## The `Formula` type
 
 The basic conceptual tool for this is the `Formula`, which has a left side and a
-right side, separated by `~`:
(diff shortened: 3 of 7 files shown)


ararslan:
Hah, so here's the thing. The curse of space-delimited macro arguments:
```julia
@model y => x1 + x2 - 1 # becomes Formula(:y, :(x1 + x2 - 1))
@model y => x1 + x2 -1  # errors, since -1 is parsed as a macro argument
```
I'm not sure what the best course of action for this is. For this approach to continue to be viable, we may have to require using parentheses, e.g.
```julia
@model(y => x1 + x2 -1) # parsed identically regardless of spacing
```
That's not ideal, but it's not _so_ bad (IMO). Thoughts?

nalimilan:
Thanks for doing this. Indeed it looks like we're going to have to recommend using the parenthesized call to avoid spacing issues.

Regarding the name of the macro, why not `@formula` instead? It returns a `Formula` object, not a `Model`. A model is more than a formula. We could introduce convenience macros later like `@fit` or `@model` which would take a formula plus a model family, a distribution, etc.

Finally, I would still use `~` rather than `=>` though, unless we're certain that it won't parse as an infix operator in the future. The tilde is a relatively established convention, while the arrow goes in the wrong direction. Or we could simply use `=`. Since that's pure bikeshedding, let's make a small poll: add thumbs up for `~`, thumbs down for `=>`, and laugh (!) for `=` (multiple choices allowed).

tkelman:
Someone will have to implement the parsing change before we can be sure, but I suspect `~` will parse as a normal infix operator just without a definition in base, until the deprecation can make its way around. What I don't know about the formula macro here is whether that can be made to work on existing Julia versions with `~` parsing as an infix macro while also inside a conventional macro.

For all but the simplest one-block annotation macros, I think parenthesizing is better form anyway.

kleinschmidt:
You could also do something like creating a new expression for the RHS is there's more than 2 arguments. Something like `Expr(:call, [:+, args[2:end]...])`.

ararslan:
Thanks for the feedback and help here, guys!

> Regarding the name of the macro, why not `@formula` instead?

I went with `@model` because there was no opposition (but also no feedback except from David Kleinschmidt). `@formula` seems fine.

> The tilde is a relatively established convention, while the arrow goes in the wrong direction

I'm not sure what you mean about the direction. Can you elaborate? I went with `=>` per an initial suggestion from @dmbates. Personally I prefer it to `~`; I've always found that convention rather unfortunate—a syntax inherited from S that seeped into other languages trying to emulate R's functionality. (I guess there's a paper somewhere that uses the notation, but IIRC it postdates its use in S? Could be wrong on that.) Though you're right, it has fairly widespread precedent and there isn't really a solid reason to deviate.

I had played around with using `=` but it makes parsing a formula out of the resulting expression more complicated when the formula itself is more complex. It's not *not* doable, just more complicated. 🙂 

> What I don't know about the formula macro here is whether that can be made to work on existing Julia versions with `~` parsing as an infix macro while also inside a conventional macro.

Should be fine, actually. Once `~` is no longer a macro, the macro here will have to change its logic (the head will no longer be `macrocall` with `@~` as its first argument, but will have `~` as the head) but that's not hard to accommodate.

> A reasonable compromise might be to do the transformation (combining all the RHS arguments with `+`) but warn people that that's ambiguous when it happens.

That seems reasonable to me. Then it's just a vararg macro, right?

kleinschmidt:
Outside of R and friends, I've seen the tilde used in a kind of descriptive way in statistics/ML papers, to describe in a more informal way the form of how a random variable depends on others. E.g., `y ~ Normal(mu, sigma^2)`, instead of writing out the PDF for the normal distribution. But maybe that's people punning on how they work in R... Regardless, in my experience it's a pretty common way of expressing the relationship between random variables (and often in the kind of high-level way that a formula is meant to capture)

ararslan:
Right, ~ is used for "distributed as" when describing the probability distribution of a random variable. But does that really translate to models? I suppose in the sense of explaining the variance of *Y* based on some *X*<sub>1</sub>, *X*<sub>2</sub>, ..., *X*<sub>*p*</sub> it sort of makes sense, as the *X*<sub>*i*</sub> have their own probability distributions. I guess I just haven't seen it for specifying models though. In my experience, that's nearly always =.

kleinschmidt:
Yes, "distributed as" is a much more concise way of saying what I was thinking. I've only ever seen `=` when it's written out very explicitly, e.g. `y_i = alpha + x_1,i beta_1 + x_2,i beta_2 + ...`.

simonster:
We definitely need to encourage parens, since `fit(Model, @model y ~ 1 + x1 + x2, ...)` is parsed as `fit(Model, @model(y ~ 1 + x1 + x2, ...))` and not `fit(Model, @model(y ~ 1 + x1 + x2), ...)` (JuliaLang/julia#12021).

ararslan:
Simon makes a good point, and I think that makes a compelling case for *requiring* parentheses here, which would also make the vararg thing unnecessary.

src/formula.jl
```diff
@@ -16,11 +16,18 @@ type Formula
...
-    formula.head === :(=>) || error("expected formula separator =>, got $(formula.head)")
-    length(formula.args) == 2 || error("malformed expression in formula")
-    lhs = Base.Meta.quot(formula.args[1])
-    rhs = Base.Meta.quot(formula.args[2])
+macro formula(ex)
+    if ex.head === :macrocall && first(ex.args) === Symbol("@~")
+        length(ex.args) == 3 || error("malformed expression in formula")
+        lhs = Base.Meta.quot(ex.args[2])
+        rhs = Base.Meta.quot(ex.args[3])
+    elseif ex.head === :(~)
```
tkelman: wouldn't it be `:call` for most infix operators?

ararslan:
Okay, now that I've fixed my dumb Vim find/replace mistakes and the tests are passing, here's what it looks like:
```julia
@formula(y ~ 1 + x1 + x2 & x3)                       # bare object
fit(SomeModel, @formula(y ~ 1 + x1 + x2 & x3), ...)  # in context
```
Does that seem reasonable?

ararslan:
Hm, once `~` is no longer a macro, we'll need to be careful about its precedence as an infix operator. If it has the same precedence as `+`, for example, things could get fairly messy with a multivariate response.

That's something else I like about `=>`; it always infix constructs a `Pair`, and you don't need to worry about its precedence with other mathematical operators. (In fact, if we were to use that, we may just be able to replace `Formula` with `Pair`...) I suppose `=` also has that advantage.

nalimilan:
> I'm not sure what you mean about the direction. Can you elaborate?

It's just that the *dependent* variable is on the LHS, so it feels weird that the `=>` arrow would go from it to the *independent* variables. A model consists in predicting the LHS *from* the RHS, not the other way around.

`=>` would have been interesting to avoid using a macro at all, but once we use our macro it doesn't have a clear advantage.

I kind of like `=` too, but outside of linear regression this notation is kind of abusive: with e.g. logistic regression or survival models, the relationship between the LHS and the RHS is more complex than that. This broader/weaker "equivalence relation" is [one of the mathematical meanings described by Wikipedia for `~`](https://en.wikipedia.org/wiki/Tilde#Mathematics).

dmbates:
Going back to the "is distributed as" interpretation of `~` in the exchange between @kleinschmidt and @ararslan, that actually fits in extremely well with linear and generalized linear models.  A linear model is
```
𝐲 ~ Normal(𝐗β, σ²I)
```
in mathematical notation and very close to that as a Julia expression.  It may be a little too wordy to write out the expression for the linear predictor in place of 𝐗 in that expression but it definitely relates the model to the expression.

For generalized linear models it is even more meaningful.  The probability model for logistic regression is
```
𝐲 ~ Bernoulli(logit.(𝐗β))
```
using `logit.` to indicate the vectorized `logit` function.

I have seen that model written incorrectly as
```
yᵢ ~ Binomial(logit(xᵢβ) + ϵᵢ)
```
or something like that so often that I have lost track.

The point is that people want to write the model in a
```
signal + noise
```
form and have a simplified expression for the distribution of `noise`.  Subtracting the mean from a multivariate `Normal` distribution leaves you with a simpler (i.e. mean 0) multivariate `Normal` distribution.  Subtracting the mean from a multivariate Bernoulli doesn't simplify the distribution.

I think I still would be in favour of an alternative notation for the model formula but I just wanted to note that the `~` in the sense of "is distributed as" does have a connection to the model being described.

ararslan:
Great explanation, Doug. Thanks! Makes total sense.

> It's just that the *dependent* variable is on the LHS, so it feels weird that the `=>` arrow would go from it to the *independent* variables. A model consists in predicting the LHS *from* the RHS, not the other way around.

Very good point. I guess in my mind I was thinking of the use of the arrow as pointing to what we're modeling the response as a function of, so "LHS as a function of (=>) RHS."

I'm on board with `~`, but I still have concerns about its precedence as an operator once it's no longer parsed as an infix macro. Assuming it gets parsed with the same precedence as something like `+` rather than something like `&&`, I guess it could be okay to stipulate that the LHS be surrounded in parentheses in the case of a multivariate response?

nalimilan:
> I'm on board with ~, but I still have concerns about its precedence as an operator once it's no longer parsed as an infix macro. Assuming it gets parsed with the same precedence as something like + rather than something like &&, I guess it could be okay to stipulate that the LHS be surrounded in parentheses in the case of a multivariate response?

AFAICT, as Tony said, `~` is going to have the same precedence as infix operators with no assigned meanings in Base. So this should work, just like `a + b ≍ x + y` is currently parsed.

tkelman:
Precedence is set on an operator by operator basis whether or not there's a meaning in base, I don't think all currently unassigned operators are at the same precedence. Given these packages are the main user so far (even after it gets changed to parse as a normal operator), I suspect it can be given whichever precedence would be most convenient for usage here. But it may change if someone comes up with a new meaning for the operator that should have different precedence.

ararslan:
Okay, sounds good to me.

ararslan:
Any further comments or are we ready to commit to doing formulas this way?

kleinschmidt:
I think this is good to go.  Can you update the documentation to reflect these changes?

ararslan:
Ah right, documentation! I had forgotten that some people like to know how to use software. 😜 Will do. Thanks for the reminder, @kleinschmidt!

kleinschmidt:
To be perfectly honest, I only thought of it because I just wrote this documentation...

kleinschmidt:
Does anyone think it's worth saying something about _why_ we're using an explicit macro, or would that be too much information for the documentation? I don't have a good sense of whether explaining these kind of design decisions is helpful to users or just noise...

ararslan:
I think documenting the design decision could be useful so long as its separate from the usage documentation. Though following Base Julia's lead, we could leave questions of design decisions to "search the GitHub issues/PRs." 😉

kleinschmidt:
I agree with separating from usage. I'd rather be explicit about it (since a lot of what happens on github assumes a lot of context that a curious user might  not have), but it's also more work to summarize things in a concise but useful way...

ararslan:
> You may be wondering why formulas in Julia require a macro, but in R they appear "bare." R uses nonstandard evaluation, allowing the formula to remain an unevaluated object while its terms are parsed out. Julia uses a much more standard evaluation mechanism, making this impossible using normal expressions. However, Julia provides macros, which allow code to be programmatically manipulated prior to evaluation. By constructing a formula using a macro, we're able to provide convenient, R-like syntax and semantics.

?

kleinschmidt:
I might say "while in R..." (instead of "but")

nalimilan:
"R uses nonstandard evaluation" -> maybe "R supports/allows nonstandard evaluation"

ararslan:
Good suggestions, thanks! I've edited my comment. How does that look now? If it looks alright I'll stick it in a commit and send it up here.

kleinschmidt:
I might even say something like "Julia, unlike R, uses macros to explicitly indicate when code itself will be manipulated before it's evaluated", just to emphasize why this is a good idea (or at least reasonable).

ararslan:
Sorry for the delay. I've incorporated the comments in the docs regarding the reasoning behind macros for formulas. Unless there are further comments, I'll go ahead and merge this later today. I'm trying to get this good to go ASAP in light of the impending feature freeze for Base Julia v0.6, wherein we hope to stop parsing `~` as a macro (ref https://github.com/JuliaLang/julia/issues/19598).

@kleinschmidt After this is merged, is there anything else that needs to be done here before we can tag a release?

Thanks so much for your help and input, everyone!

ararslan:
Not sure what the nightly failure is about but it appears unrelated

tkelman:
Can this be merged? What deprecation strategy for the parsing of `~` in base would make the migration doable soon?

ararslan:
I tried to write the macro in such a way that it should work regardless of whether `~` is a macro call or a regular call. The only thing that needs to be preserved is the parsing precedence of `~`.

tkelman:
Kinda. Are other packages using the old syntax in tests, or does this only ever appear in user code?

ararslan:
I think many packages that currently depend on DataFrames (which still defines this syntax) do use formulas in tests. Examples off the top of my head include MixedModels and FixedEffectModels.

tkelman:
And they're getting the definition from here? So I guess this package should give a dep warn for tilde when it gets called as a macro rather than inside of one?

ararslan:
This package isn't registered yet, so the definition _will_ come from here, but doesn't yet. The next release of DataFrames will not contain any of the formula code; that will all come from here. So for 0.6 compatibility with the current DataFrames we'd have to do a patch release with a deprecation warning. I guess one way to do it would be to have DataFrames define `@~` with the dep warn and things as-is here.

tkelman:
Oh! Thanks for the explanation, didn't realize that. Is the old release DataFrames branch the only place that defined a ~ macro implementation? Would this be ready for other packages to use as a replacement soon?

Or instead of DataFrames, maybe Base could keep ~ parsing as a macro but always throwing a depwarn? Looks like this may not change in Base for 0.6, but if packages can be made ready then maybe we hard change the parsing during 1.0-dev and packages might not notice if they've transitioned to this?

ararslan:
> Is the old release DataFrames branch the only place that defined a ~ macro implementation?

Should still be on master as well until this is registered, but otherwise yes.

> Would this be ready for other packages to use as a replacement soon?

I _think_ so, but @kleinschmidt may know better than I would.

> Or instead of DataFrames, maybe Base could keep ~ parsing as a macro but always throwing a depwarn? Looks like this may not change in Base for 0.6, but if packages can be made ready then maybe we hard change the parsing during 1.0-dev and packages might not notice if they've transitioned to this?

👍 Sounds like the best course of action.

PR #10, opened 2016-12-22
Dropterm

dmbates:
I think it would help to have functions for manipulating formulas.  `dropterm` is easy to define.  Methods could be added for a collection of `Symbol`s or expressions.  The reason I use this is to perform an analysis of deviance on a `GeneralizedLinearModel` (i.e. fit the original model, drop a term and fit the null model, evaluate the change in deviance).

### DIFF
--- a/src/formula.jl
+++ b/src/formula.jl
@@ -214,6 +214,11 @@ function Formula(t::Terms)
     Formula(lhs,rhs)
 end
 
+function Base.copy(f::Formula)
+    lhs = isa(f.lhs, Symbol) ? f.lhs : copy(f.lhs)
+    return Formula(lhs, copy(f.rhs))
+end
+
 """
     dropterm(f::Formula, trm::Symbol)
 
@@ -222,14 +227,26 @@ Return a copy of `f` without the term `trm`.
 # Examples
 ```jl
 julia> dropterm(foo ~ 1 + bar + baz, :bar)
-foo ~ 1 + baz
+Formula: foo ~ 1 + baz
+
+julia> dropterm(foo ~ 1 + bar + baz, 1)
+Formula: foo ~ 0 + bar + baz
 ```
 """
-function dropterm(f::Formula, trm::Symbol)
-    rhs = copy(f.rhs)
-    args = rhs.args
-    if !(Meta.isexpr(rhs, :call) && args[1] == :+ && (tpos = findlast(args, trm)) > 0)
-        throw(ArgumentError("$trm is not a summand in `$rhs`"))
+dropterm(f::Formula, trm::Union{Number, Symbol, Expr}) = dropterm!(copy(f), trm)
+
+function dropterm!(f::Formula, trm::Union{Number, Symbol, Expr})
+    rhs = f.rhs
+    if !(Meta.isexpr(rhs, :call) && rhs.args[1] == :+ && (tpos = findlast(rhs.args, trm)) > 0)
+        throw(ArgumentError("$trm is not a summand of '$(f.rhs)'"))
     end
-    Formula(f.lhs, Expr(:call, :+, deleteat!(args, [1, tpos])...))
+    if isa(trm, Number)
+        if trm ≠ one(trm)
+            throw(ArgumentError("Cannot drop $trm from a formula"))
+        end
+        rhs.args[tpos] = 0
+    else
+        deleteat!(rhs.args, [tpos])
+    end
+    return f
 end
--- a/src/formula.jl
+++ b/src/formula.jl
@@ -229,10 +229,10 @@ Return a copy of `f` without the term `trm`.
 
 # Examples
 ```jl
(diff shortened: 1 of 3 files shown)


src/formula.jl
```diff
@@ -213,3 +213,23 @@ function Formula(t::Terms)
...
+
+Return a copy of `f` without the term `trm`.
+
+# Examples
+```jl
+julia> dropterm(foo ~ 1 + bar + baz, :bar)
+foo ~ 1 + baz
+```
+"""
+function dropterm(f::Formula, trm::Symbol)
```
kleinschmidt: Terms can be expressions, too (interactions terms). I think `Union{Expr,Symbol}` might do it here.

src/formula.jl
```diff
@@ -213,3 +213,23 @@ function Formula(t::Terms)
...
+    rhs = copy(f.rhs)
+    args = rhs.args
+    if !(Meta.isexpr(rhs, :call) && args[1] == :+ && (tpos = findlast(args, trm)) > 0)
```
kleinschmidt: What about formulas with only one term, like `y ~ a`?

ararslan:
It might be nice to provide a way to drop the intercept as well using this function. What would that look like? `dropterm(y ~ 1 + x, 1)`?

ararslan:
Also, would it be worth it to provide `dropterm!` that modifies the `Formula` in place, with `dropterm(f, t) = dropterm!(copy(f), t)`?

src/formula.jl
dmbates: I would like to discourage formulas like `y ~ a` and require either `y ~ 1 + a` or `y ~ 0 + a`

Also, what should `dropterm(y ~ a, :a)` return?  The lhs of a formula can be `nothing` but I don't know how to make sense of a rhs of `nothing`.

kleinschmidt: Yeah, that depends on the implicit intercept behavior. At the moment, I'd say it should return `y ~ 1` (e.g., the null model).

The `y ~ 0 + a` case is tricky, too; can you even evaluate `y ~ 0`?

I suppose you could just check whether the RHS would be empty and create `y ~ 1` for now, and we can fix it when we change intercepts to always be explicit.

ararslan: IMO `dropterm(y ~ a, :a)` should be an error

dmbates:
@ararslan I agree that `dropterm!` should be the base definition and `dropterm` should call it.

Regarding dropping the intercept, that should be its own method (signature of `(::Formula, ::Integer)`) so that `dropterm(y ~ 1 + x, 1)` returns `y ~ 0 + x`, not `y ~ +x`.  An interesting question is what should `dropterm(y ~ 0 + x)` return?  It might return `y ~ 1 + x` or it might be cleaner to throw an error.

ararslan:
> An interesting question is what should `dropterm(y ~ 0 + x)` return?

Are you missing an argument there? Do you mean `dropterm(y ~ 0 + x, 0)` or `dropterm(y ~ 0 + x, :x)`?

dmbates:
@ararslan Yes, I meant "what should `dropterm(y ~ 0 + x, 0)` return?"

ararslan:
Oh, okay. I would say that should be an error, since you're dropping a lack of intercept, which itself is a bizarre concept.

dmbates:
@kleinschmidt @ararslan I have added a `dropterm!`, which `dropterm` now calls, and extended the signature of the argument to `trm::Union{Number, Symbol, Expr}`  I thinkt the CI failures for the nightly version are unrelated to these changes.

test/formula.jl
```diff
@@ -106,4 +106,9 @@ t = Terms(y ~ x1 * x2 * x3)
 ## @test t.terms == [:x2, :(x1&x2)]    # == [:(1 & x1)]
 ## @test t.eterms == [:y, :x1, :x2]
 
+@test dropterm(foo ~ 1 + bar + baz, :bar).rhs == :(1 + baz)
+@test dropterm(foo ~ 1 + bar + baz, 1).rhs == :(0 + bar + baz)
+@test_throws ArgumentError dropterm(foo ~ 0 + bar + baz, 0)
+@test_throws ArgumentError dropterm(foo ~ 0 + bar + baz, :boz)
+
```
ararslan: Should test `dropterm!` as well. Maybe ensure that the input formula is indeed modified.

ararslan:
I added a few commits to update this branch to accommodate the change to the `@formula` replacement for `Base.@~`. That should get the tests passing. LGTM, pending green CI.

dmbates:
Thanks for updating this pull request @ararslan

ararslan:
Hm, just realized that the docstring should have made its way into the documentation somewhere by adding it to one of the markdown files. Not really sure what the best section to put it in would be though.

dmbates:
The `Formula` section of the `docs/src/formula.md`?

PR #11, opened 2017-01-05
Update Terms for Julia 0.6 changes

ararslan:
This does a couple of things:
* `tt` is inferred as `Array{Any,1}`, as is `tt .== 1`, so `!(tt .== 1)` fails. Changed to `filter!`.
* Vectorized `|` is deprecated and `.|` doesn't exist in 0.5. Changed to `map`.

### DIFF
--- a/src/formula.jl
+++ b/src/formula.jl
@@ -176,8 +176,8 @@ evt(a) = Any[a]
 function Terms(f::Formula)
     rhs = condense(distribute(dospecials(f.rhs)))
     tt = unique(getterms(rhs))
-    tt = tt[!(tt .== 1)]             # drop any explicit 1's
-    noint = (tt .== 0) | (tt .== -1) # should also handle :(-(expr,1))
+    filter!(t -> t != 1, tt)                          # drop any explicit 1's
+    noint = BitArray(map(t -> t == 0 || t == -1, tt)) # should also handle :(-(expr,1))
     tt = tt[!noint]
     oo = Int[ord(t) for t in tt]     # orders of interaction terms
     if !issorted(oo)                 # sort terms by increasing order


src/formula.jl
```diff
@@ -176,8 +176,8 @@ evt(a) = Any[a]
 function Terms(f::Formula)
     rhs = condense(distribute(dospecials(f.rhs)))
     tt = unique(getterms(rhs))
-    tt = tt[!(tt .== 1)]             # drop any explicit 1's
-    noint = (tt .== 0) | (tt .== -1) # should also handle :(-(expr,1))
+    filter!(t -> t != 1, tt)                          # drop any explicit 1's
+    noint = BitArray(map(t -> t == 0 || t == -1, tt)) # should also handle :(-(expr,1))
```
andreasnoack: `bitbroadcast`

ararslan: Looks nifty but is apparently deprecated in 0.6

andreasnoack: Ah. So `broadcast` (but not `map`) will figure out that the return array should be `BitArray`.

ararslan: That will still fail the currently failing test because `broadcast(t -> t == 1, tt)` doesn't produce an empty `BitArray` if `tt` is empty; it produces `Array{Any,1}`.

ararslan:
Any objection to merging this as-is, @andreasnoack?

ararslan:
I'll take that as a "no" 😄

src/formula.jl
nalimilan: I think https://github.com/JuliaLang/julia/pull/19854 fixed this. On latest master:
```julia
julia> broadcast(t -> t == 1, Float64[])
0-element BitArray{1}

```

PR #13, opened 2017-02-18
Get tests pass again, fix handling of string columns

nalimilan:
Move to DataTables temporarily since the code is written to work with Nullable for now. This allows porting https://github.com/JuliaStats/DataFrames.jl/pull/1147/commits/781bbbd0e6724f697fd2dc35867961df548f4434 to StatsModels (first commit), which is a first step towards supporting any kind of categorical input (i.e. both `DataArray` and `CategoricalArray`).

If this works, we could remove the conflicting API from DataTables so that we no longer need the `using` hacks. Then we could start relying only on the AbstractTable interface.

### DIFF
--- a/src/modelframe.jl
+++ b/src/modelframe.jl
@@ -137,7 +137,7 @@ end
 
 ## Default NULL handler.  Others can be added as keyword arguments
 function null_omit(df::DataTable)
-    cc = complete_cases(df)
+    cc = completecases(df)
     df[cc,:], cc
 end
 
--- a/src/modelframe.jl
+++ b/src/modelframe.jl
@@ -1,5 +1,5 @@
 """
-Wrapper which combines Formula (Terms) and an AbstractDataFrame
+Wrapper which combines Formula (Terms) and an AbstractDataTable
 
 This wrapper encapsulates all the information that's required to transform data
 of the same structure as the wrapped data frame into a model matrix.  This goes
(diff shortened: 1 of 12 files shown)


src/modelframe.jl
```diff
@@ -102,34 +104,48 @@ end
 
 const DEFAULT_CONTRASTS = DummyCoding
 
+_levels(x::Union{CategoricalArray, NullableCategoricalArray}) = levels(x)
```
nalimilan: Note to self: need to change this to `unique`, cf. https://github.com/JuliaStats/DataFrames.jl/pull/1147#discussion_r96266095.

README.md
```diff
@@ -9,4 +9,4 @@ in Julia.
 
 **Documentation:** [![Latest](https://img.shields.io/badge/docs-latest-blue.svg)](https://juliastats.github.io/StatsModels.jl/latest)
 
-Originally part of [DataFrames.jl](http://github.com/JuliaStats/DataFrames.jl).
+Originally part of [DataTables.jl](http://github.com/JuliaStats/DataTables.jl).
```
tkelman: the old one is a bit more accurate

nalimilan: Technically both are true, but indeed that's not the most informative choice...

src/modelframe.jl
nalimilan: Actually that was quite easy. What is going to be harder is to handle DataArrays the same way, since `unique(::PooledDataArray)` does not return values in the order specified by levels. Since `unique` does not guarantee any particular order for levels, I suggest we change DataArrays to follow the ordering of levels (just like `CategoricalArray`). That way, we don't need any common API other than `unique`.

nalimilan:
Julia 0.6 failure is fixed by https://github.com/JuliaLang/julia/pull/20665.

src/modelframe.jl
nalimilan: See https://github.com/JuliaStats/DataArrays.jl/pull/237.

nalimilan:
I've squashed some commits and removed the hacks now that DataTables no longer export conflicting function definitions. So I'd say it's good to go.

README.md
kleinschmidt: It might be worth expanding on this a little in the readme; something along the lines of "this was once part of dataframes, now it depends on datatables, but eventually will support generic tabular data stores including both dataframes and datatables"

kleinschmidt: (Just to avoid confusion when people come here looking for help/information)

nalimilan: Yes. I would wait until things have stabilized a bit more, and then explain how the package is supposed to be used. Anyway we need generic interfaces very soon to support DataFrames.

PR #15, opened 2017-02-22
Remove some uses of CategoricalArrays, update docs

nalimilan:
The first commit fixes docs which were not updated by https://github.com/JuliaStats/StatsModels.jl/pull/13.

The second one removes special cases for `CategoricalArray` which are no longer really needed.

### DIFF
--- a/src/modelframe.jl
+++ b/src/modelframe.jl
@@ -141,14 +141,10 @@ function null_omit(df::DataTable)
     df[cc,:], cc
 end
 
-_droplevels!(x::Any) = x
-_droplevels!(x::Union{CategoricalArray, NullableCategoricalArray}) = droplevels!(x)
-
 function ModelFrame(trms::Terms, d::AbstractDataTable;
                     contrasts::Dict = Dict())
     df, msng = null_omit(DataTable(map(x -> d[x], trms.eterms)))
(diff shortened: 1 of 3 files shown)


src/modelframe.jl
```diff
@@ -141,14 +141,10 @@ function null_omit(df::DataTable)
...
 end
 
-_droplevels!(x::Any) = x
-_droplevels!(x::Union{CategoricalArray, NullableCategoricalArray}) = droplevels!(x)
-
 function ModelFrame(trms::Terms, d::AbstractDataTable;
                     contrasts::Dict = Dict())
     df, msng = null_omit(DataTable(map(x -> d[x], trms.eterms)))
     names!(df, convert(Vector{Symbol}, map(string, trms.eterms)))
-    for c in eachcol(df) _droplevels!(c[2]) end
```
nalimilan: @kleinschmidt Can you confirm this is OK? I've checked that unused levels continue to be skipped (since we call  `unique`), but I'd like to be sure I'm not missing something.

kleinschmidt: LGTM. This is covered by tests: https://github.com/JuliaStats/StatsModels.jl/blob/master/test/modelmatrix.jl#L275

Issue #14, opened 2017-02-22
Avoid hard-coding for concrete implementations


andreasnoack:
I'm wondering to what extent it would be practical to avoid hard-coding for specific implementations in this package. In particular for NullableArrays and DataTables since I'd like to use the functionality here with DataFrames and DataArrays. Commenting out specific uses of NullableArrays and removing the uses of `DataTable` signatures (could be some kind of AbstractTable in the future) made some of the basic functionality work. Some definitions like https://github.com/JuliaStats/StatsModels.jl/blob/master/src/modelframe.jl#L108 seem misplaced here in any case. I believe they should be defined in their home packages. @nalimilan why do you define the underscore versions of these functions? Other uses might be harder to get rid of. In particular the use of the Categorical types. I wondered if we could define an abstract categorical which also PooledDataArrays could be subtypes of.

A very different approach could be to use https://github.com/davidagold/Relations.jl/ as the unifying framework here. I think it is very interesting and might be a way to go beyond in-memory data frame representations. However, there are probably a lot to sort out before could consider such a change whereas the changes discussed above are pretty simple.

nalimilan:
The `_unique` methods are needed to unwrap nullables. Fortunately, special-casing `Nullable` does not interfere at all with `DataArray` support, which does not need unwrapping. Since this code is working on complete cases, it will never see a null or a `NA` and should work fine with either. See also https://github.com/JuliaStats/DataArrays.jl/pull/237.

Actually, it won't be too hard to stop hard-coding `CategoricalArray`. In https://github.com/JuliaStats/StatsModels.jl/pull/15, I remove half of the uses, which are no longer needed. As you suggest, it would make sense to define `AbstractCategoricalArray` (probably in `AbstractTables` again), since the remaining occurrence of `CategoricalArray` in the code base is for `modelmat_cols`, where it's much more efficient to use the integer codes than to compare strings one by one to get the index of the column they correspond to. That's pretty easy change to make, with the slight complication that we also need `AbstractNullableCategoricalArray`, which won't be used by `PooledDataArray` since no unwrapping is required for it.

Then the other easy step is to replace `AbstractDataTable` with `AbstractTable` and make `DataFrame` and `DataTable` inherit from the latter.

Longer-term, I think it would indeed be great to move to a tuple-based approach like Relations.jl, which would avoid lots of copies and allow supporting any kind of source. But for now we don't really need it to support both DataFrames and DataTables.

[...]

PR #16, opened 2017-02-24
WIP: First steps in defining a common interface for Tables/Frames

andreasnoack:
There is almost nothing here yet. Just wanted to open the PR such that we have a place to discuss details of #14. This will require a new package with the common interface. So far I've named it `TableBase`. Both `DataFrames` and `DataTables` should then become subtypes of this new `Table` and some subset of the functions in the two packages should be defined as erroring methods in `TableBase` and overloaded in `DataFrames` and `DataTables`.

### DIFF
--- a/docs/src/formula.md
+++ b/docs/src/formula.md
@@ -14,7 +14,7 @@ fields with possibly heterogeneous types.  One of the primary goals of
 `StatsModels` is to make it simpler to transform tabular data into matrix format
 suitable for statistical modeling.
 
-At the moment, "tabular data" means an `AbstractDataTable`.  Ultimately, the
+At the moment, "tabular data" means an `Table`.  Ultimately, the
 goal is to support any tabular data format that adheres to a minimal API,
 **regardless of backend**.
 
@@ -88,7 +88,7 @@ dropterm
 
 The main use of `Formula`s is for fitting statistical models based on tabular
 data.  From the user's perspective, this is done by `fit` methods that take a
-`Formula` and a `DataTable` instead of numeric matrices.
+`Formula` and a `Table` instead of numeric matrices.
 
 Internally, this is accomplished in three stages:
 
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -21,5 +21,6 @@ developers when dealing with statistical models and tabular data.
     * `RegressionModel`
 
 Much of this package was formerly part
-of [`DataTables`](https://www.github.com/JuliaStats/DataTables.jl)
+of [`DataTables`](https://www.github.com/JuliaStats/DataTables.jl)/
+[`DataFrames`](https://www.github.com/JuliaStats/DataFrames.jl)
 and [`StatsBase`](https://www.github.com/JuliaStats/StatsBase.jl).
--- a/src/StatsModels.jl
+++ b/src/StatsModels.jl
@@ -3,10 +3,12 @@ __precompile__(true)
 module StatsModels
 
 using Compat
-using DataTables
+using TableBase
+using TableBase: Table
+# using DataTables
 using StatsBase
-using NullableArrays
-using CategoricalArrays
+# using NullableArrays
+# using CategoricalArrays
 
 
 export @formula,
@@ -23,6 +25,10 @@ export @formula,
        coefnames,
        dropterm,
        setcontrasts!
+
+# TEMPORARY DEFINITIONS
+const AbstractCategoricalVector = Any
+
 
 map(include,
     [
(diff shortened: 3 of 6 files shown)


nalimilan:
Why not call this `AbstractTable` in the end? It find it more explicit, and that's the name we've been using for a long time now.

andreasnoack:
That was also my first idea but I realized that @davidagold had added developed AbstractTable in https://github.com/davidagold/AbstractTables.jl with a dependency on `NullableArrays` and `StructuredQueries` which seemed different from the `XBase` like package needed here. If I'm wrong and the dependency on NullableArrays isn't critical then it is just a search replace change. The main task for now is to figure out how to write the methods such that they avoid details of concrete implementations.

docs/src/formula.md
```diff
@@ -14,7 +14,7 @@ fields with possibly heterogeneous types.  One of the primary goals of
 `StatsModels` is to make it simpler to transform tabular data into matrix format
 suitable for statistical modeling.
 
-At the moment, "tabular data" means an `AbstractDataTable`.  Ultimately, the
+At the moment, "tabular data" means an `Table`.  Ultimately, the
```
kleinschmidt: a `Table`, not an

nalimilan:
We would need to ask David, but I don't think we want to provide different table abstractions. And there's only [one use](https://github.com/davidagold/AbstractTables.jl/blob/4de25b55271fe8f208d85fcb0100981223e9f043/src/column_indexable/query/summarize.jl#L14) of `NullableArray` in the `src/` dir. The described interface is very general (and he said it wasn't completely settled yet).

docs/src/formula.md
nalimilan: ...or change `Table` to `AbstractTable`. :-)

daemonomania:
So, this is David, who has permanently locked himself out of his davidagold GH account. Alas.

I think it should be fine to remove those dependencies from AbstractTables. I'd be hesitant, however, to start changing all the method signatures in this PR from `DataTable` to (Abstract)`Table`, since most of them do make assumptions about the underlying storage pattern. It will probably be more useful, as we think about abstractions, to have `DataTable` in these signatures so it's immediately obvious which methods make such assumptions, and we're not trying to recover that information each time we look at the method.

andreasnoack:
The idea here is to remove `DataTables` as a dependency which requires that `DataTable` is not in the signature. If methods in this package require a specific storage then we should in the longer run try to loosen such assumptions. In the short run, we'd have to make these assumptions for the abstract type. If that is too strong for `AbstractTable` then it could either be a subtype of `AbstractTable` or a completely new type.

daemonomania:
Seems like traits would be useful here.

quinnj:
@andreasnoack, where is the current TableBase code? I'm starting to actively compile the various "abstracttable" attempts into a single package (taking parts of DataStreams, the AbstractDataFrame code, David's AbstractTables as well as his Relations.jl package).

For everyone else, I think the biggest thing that would help as we try to converge on a single "AbstractTable" interface is the strongest set of use-cases for an AbstractTable. I'm going to try and dig more into the code here in StatsModels to figure out what exactly it "needs" from an AbstractTable interface, but it'd be great to have other strong use-cases. For me, I'm coming from the context of DataStreams, which requires functionality like getting the "schema" of a table and being able to get/set individual cells, as well as entire columns at a time. I think the ideal goal is to come up with:

* The minimum set of required and optional methods a table implementation would need to implement
* The "interface" functionality that would be fully provided (and cover all necessary use-cases) once a table has implemented the interface

In terms of a starter list of use-cases to consider, I can think of:

* DataStreams: generalizing tabular data IO across storage/transport formats
* StatsModels: encoding data + terms + formulas in a representation amenable to various statistics operations
* Query/StructuredQueries: performing SQL-like "get" operations on tabular data; selecting, filtering, joining, grouping, etc.
* Other data processing packages? I'm thinking like Clustering.jl, which could potentially work directly on any table-type, though I'm less clear on how generally applicable this might be because different processing libraries might require more specific data manipulation routines.

nalimilan:
I think in theory the required interface for StatsModels is quite limited:
- get the number and types of the variables
- get the number of observations
- skip observations with missing values in one of the (selected) variables
- optionally get the ordering of levels for categorical variables (and possibly the name of the reference level, so that we don't necessarily assume it's the first one)
- access the values of cells one by one (to fill the model matrix); variable/column-wise would be faster but row-wise would work too

The issue is that the current code makes stronger assumptions about being able to access columns as vectors. I think @kleinschmidt had plans to change this, but it will take some work.

So in the short term I think we should just write an abstraction which specifies that the abstract data table (or whatever we call that particular interface/trait) can be indexed with variable names, as that it returns vectors. We can always remove that requirement later, or provide inefficient fallbacks which would create such vectors for data tables which use a different storage.

See also https://github.com/JuliaStats/StatsModels.jl/issues/14.

kleinschmidt:
Superseded by #71

PR #22, opened 2017-05-16
Small depreciation fix.

Nosferican:
!(B::BitArray) is deprecated, use .!(B) instead.


src/formula.jl
```diff
@@ -181,7 +181,7 @@ function Terms(f::Formula)
     tt = unique(getterms(rhs))
     filter!(t -> t != 1, tt)                          # drop any explicit 1's
     noint = BitArray(map(t -> t == 0 || t == -1, tt)) # should also handle :(-(expr,1))
-    tt = tt[!noint]
+    tt = tt[.!noint]
```
ararslan: This syntax only works on Julia 0.6 and later. Since this package still supports 0.5, this should be `map(!, noint)` instead.

PR #26, opened 2017-06-13
hash and equality for ContrastsMatrix types

kleinschmidt:
Allows ContrastsMatrices to be compared (and used as keys in dicts etc.)

### DIFF
--- a/src/contrasts.jl
+++ b/src/contrasts.jl
@@ -89,6 +89,17 @@ type ContrastsMatrix{C <: AbstractContrasts, T}
     contrasts::C
 end
 
+# only check equality of matrix, termnames, and levels, and that the type is the
+# same for the contrasts (values are irrelevant).  This ensures that the two
+# will behave identically in creating modelmatrix columns
+Base.:(==){C<:AbstractContrasts,T}(a::ContrastsMatrix{C,T}, b::ContrastsMatrix{C,T}) =
+    a.matrix == b.matrix &&
+    a.termnames == b.termnames &&
+    a.levels == b.levels
+
+Base.hash{C}(a::ContrastsMatrix{C}, h::UInt) =
+    hash(C, hash(a.matrix, hash(a.termnames, hash(a.levels, h))))
+
 """
 An instantiation of a contrast coding system for particular levels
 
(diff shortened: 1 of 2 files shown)


ararslan:
Cool! Could you add some tests for this?

src/contrasts.jl
```diff
@@ -89,6 +89,17 @@ type ContrastsMatrix{C <: AbstractContrasts, T}
     contrasts::C
 end
 
+# only check equality of matrix, termnames, and levels, and that the type is the
+# same for the contrasts (values are irrelevant).  This ensures that the two
+# will behave identically in creating modelmatrix columns
+Base.:(==){C<:AbstractContrasts,T}(a::ContrastsMatrix{C,T}, b::ContrastsMatrix{C,T}) =
```
nalimilan: Shouldn't there be a method for when `C` differs between matrices? You could simply broaden the signature, and add another condition `C === C` in the body.

kleinschmidt: I think that case is captured by the default fallback.

kleinschmidt:
I was a little too quick in my initial PR. I've added tests for a bunch of cases where different ways of constructing contrasts should lead to the same result (which should help the robustness of the package overall)

src/contrasts.jl
nalimilan: Ah, right. Though it would be clearer with `C1 === C2`, and maybe even shorter since the signature wouldn't need type parameters.

test/contrasts.jl
```diff
@@ -5,11 +5,36 @@ using DataTables
 using CategoricalArrays
 using StatsModels
 
+import StatsModels: ContrastsMatrix
```
nalimilan: Very minor point, but `using` is enough here.

src/contrasts.jl
kleinschmidt: I agree it might be clearer but how would that get rid of the type parameters? We'd still need them to be able to compare the two contrasts types

test/contrasts.jl
nalimilan: But it works too. `import` is for extending functions.

kleinschmidt: Hm. It doesn't look like it works without the `import`; the tests fail with `UndefVarError: ContratsMatrix not defined`

kleinschmidt: Oh I see what you mean now: `using StatsModels: ContrastsMatrix` is sufficient.

PR #28, opened 2017-06-27
Add model_response methods

### LINKED ISSUE #27: Extend StatsBase.model_response methods with a Distribution argument
I have been thinking about conversion of a binary `PooledDataArray` or `CategoricalVector` to a 0/1 floating point vector in the `StatsBase.model_response` method.  The trick is in determining what kind of response vector (or matrix or other structure) is desired.  It occurred to me that the Distribution type in the model tells us that.  For an `lm` or, in my case, `lmm` model the distribution type is Normal.  For a `glm` or `glmm` model the distribution type is given in the model specification.

In creating a `GLM.GlmResp` object the response vector is checked to ensure that all values are in the support of the distribution so we might as well use that as the guide.

If this seems reasonable I will create a pull request.

Is it worthwhile changing the name of this extractor at the same time?  The name `model_response` was copied from R and is not Julianesque.

dmbates:
- add `model_response` methods for `Bernoulli` distribution
- also cleaned up deprecated `Array` call signatures in `droprandomeffects`

Fixes #27.

### DIFF
--- a/src/modelframe.jl
+++ b/src/modelframe.jl
@@ -172,19 +172,40 @@ function setcontrasts!(mf::ModelFrame, new_contrasts::Dict)
 end
 setcontrasts!(mf::ModelFrame; kwargs...) = setcontrasts!(mf, Dict(kwargs))
 
+convertinsupport(v::AbstractVector, d::Distribution) = Vector{partype(d)}(v)
+
+function convertinsupport(v::Union{AbstractCategoricalVector, AbstractNullableCategoricalVector},
+    d::Bernoulli)
+    vdrp = droplevels!(copy(v))
+    levs = CategoricalArrays.levels(vdrp)
+    llv = length(levs)
+    if llv < 2
+        zeros(vdrp, partype(d))
+    elseif llv == 2
+        partype(d)[get(x) == levs[2] for x in vdrp]
+    else
+        throw(ArgumentError("length(levels(droplevels!(v))) = $llv, should be ≤ 2"))
+    end
+end
+
+function convertinsupport(v::Vector{String}, d::Bernoulli)
+    convertinsupport(CategoricalVector(v), d)
+end
+
 """
     StatsBase.model_response(mf::ModelFrame)
 Extract the response column, if present.  `DataVector` or
 `PooledDataVector` columns are converted to `Array`s
 """
-function StatsBase.model_response(mf::ModelFrame)
+function StatsBase.model_response(mf::ModelFrame, d::Distribution)
     if mf.terms.response
-        convert(Array, mf.df[mf.terms.eterms[1]])
+        convertinsupport(mf.df[mf.terms.eterms[1]], d)
     else
         error("Model formula one-sided")
     end
 end
 
+StatsBase.model_response(mf::ModelFrame) = model_response(mf, Normal())
 
 """
     termnames(term::Symbol, col)
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,6 @@
 julia 0.5
 DataTables
+Distributions
 CategoricalArrays 0.0.6
 NullableArrays 0.0.10
 StatsBase 0.11.1
--- a/src/StatsModels.jl
+++ b/src/StatsModels.jl
@@ -4,6 +4,7 @@ module StatsModels
 
 using Compat
 using DataTables
+using Distributions
 using StatsBase
 using NullableArrays
 using CategoricalArrays
--- a/src/modelmatrix.jl
+++ b/src/modelmatrix.jl
@@ -98,7 +98,7 @@ function droprandomeffects(trms::Terms)
     if !any(retrms)  # return trms unchanged
         trms
     elseif all(retrms) && !trms.response   # return an empty Terms object
-        Terms(Any[],Any[],Array(Bool, (0,0)),Array(Bool, (0,0)), Int[], false, trms.intercept)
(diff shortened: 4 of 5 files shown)


src/modelframe.jl
```diff
@@ -172,19 +172,40 @@ function setcontrasts!(mf::ModelFrame, new_contrasts::Dict)
 end
 setcontrasts!(mf::ModelFrame; kwargs...) = setcontrasts!(mf, Dict(kwargs))
 
+convertinsupport(v::AbstractVector, d::Distribution) = Vector{partype(d)}(v)
+
+function convertinsupport(v::Union{AbstractCategoricalVector, AbstractNullableCategoricalVector},
+    d::Bernoulli)
```
nalimilan: Wrong indentation.

src/modelframe.jl
```diff
@@ -172,19 +172,40 @@ function setcontrasts!(mf::ModelFrame, new_contrasts::Dict)
...
+        throw(ArgumentError("length(levels(droplevels!(v))) = $llv, should be ≤ 2"))
+    end
+end
+
+function convertinsupport(v::Vector{String}, d::Bernoulli)
+    convertinsupport(CategoricalVector(v), d)
+end
+
 """
     StatsBase.model_response(mf::ModelFrame)
```
nalimilan: Should update the docstring.

src/modelframe.jl
```diff
@@ -172,19 +172,40 @@ function setcontrasts!(mf::ModelFrame, new_contrasts::Dict)
...
+function StatsBase.model_response(mf::ModelFrame, d::Distribution)
     if mf.terms.response
-        convert(Array, mf.df[mf.terms.eterms[1]])
+        convertinsupport(mf.df[mf.terms.eterms[1]], d)
     else
         error("Model formula one-sided")
     end
 end
 
+StatsBase.model_response(mf::ModelFrame) = model_response(mf, Normal())
```
nalimilan: Just use `d=Normal()` in the signature of the previous method.

src/modelframe.jl
```diff
@@ -172,19 +172,40 @@ function setcontrasts!(mf::ModelFrame, new_contrasts::Dict)
 end
 setcontrasts!(mf::ModelFrame; kwargs...) = setcontrasts!(mf, Dict(kwargs))
 
+convertinsupport(v::AbstractVector, d::Distribution) = Vector{partype(d)}(v)
+
+function convertinsupport(v::Union{AbstractCategoricalVector, AbstractNullableCategoricalVector},
+    d::Bernoulli)
+    vdrp = droplevels!(copy(v))
```
nalimilan: Rather than allocating a copy of `v`, would be more efficient to do `levs = intersect(levels(v), unique(v))`. (I should add a `drop` argument to `levels` which would do exactly that.)

src/modelframe.jl
```diff
@@ -172,19 +172,40 @@ function setcontrasts!(mf::ModelFrame, new_contrasts::Dict)
...
+    levs = CategoricalArrays.levels(vdrp)
+    llv = length(levs)
+    if llv < 2
+        zeros(vdrp, partype(d))
+    elseif llv == 2
+        partype(d)[get(x) == levs[2] for x in vdrp]
```
nalimilan: `get` won't work for non-nullable arrays. If you want the most efficient code, that would be:
```julia
x = get(v.pool, levs[2])
partype(d)[ref == x for ref in v.refs]
```

src/modelframe.jl
```diff
@@ -172,19 +172,40 @@ function setcontrasts!(mf::ModelFrame, new_contrasts::Dict)
...
+    else
+        throw(ArgumentError("length(levels(droplevels!(v))) = $llv, should be ≤ 2"))
```
nalimilan: Maybe the message should rather mention the Bernouilli distribution and the number of levels present in the response variable? If people fitting a GLM see this without calling `convertinsupport` themselves, they don't know what is `v`.

src/modelframe.jl
```diff
@@ -172,19 +172,40 @@ function setcontrasts!(mf::ModelFrame, new_contrasts::Dict)
...
+    end
+end
+
+function convertinsupport(v::Vector{String}, d::Bernoulli)
```
nalimilan: You'll need a method for `AbstractVector{Nullable{String}}` too. And this one should be made more general, accepting any `AbstractVector{String}`.

Nosferican:
This should probably be adapted to `response` and StatsBase changed `model_response` to `response`. `model_response(::ModelFrame)` should be adjusted accordingly as well.

haberdashPI:
Just ran into the absence of this feature: looks like this is pretty old at this point. Is there a plan to incorporate it? Doesn't look too hard to finish this up from what I see here. I might be able to put some time into it if that makes sense given current plans.

Also, it would be great if, until this feature exists, there is a better error message: the error you get when using a categorical variable and try to run GLM is very confusing (something about a missing method supporting Array{T,2} for `y`). I'd be happy to at least do that.

kleinschmidt:
I'm going to close this, for a few reasons.  First, IIRC it's pre- #71 and hence would require MAJOR refactoring if it's going to be useful as-is.  Second, I think this feature is better implemented in a specialized package like GLM, both because we've recently dropped the dependency on Distributions.jl (#195), and also because it's not immediately obvious how to implement the function here as a stand-alone term that would be useful (since in GLM it's not the *outcome* itself that's transformed but actually the linear predictor)

PR #30, opened 2017-09-02
Move back to DataFrames and port from Nullable to Missings

nalimilan:
Keep Nullable in cases where it's used to represent "software engineer's nulls",
i.e. not missing values in data, since the replacement for that in Base is not yet clear.

It should also be possible to avoid returning a `Union{T, Null}` vector from `predict()` when
the input variables are not nullable by storing that in the type information, but
this is left for later.

Also require Julia 0.6.

----------

The two commits should be merged as the first one does not make sense in isolation. Tests won't pass until a new DataFrames release based on `Union{Null, T}` is tagged (tentatively referred to as 0.10 here).

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.6
 DataFrames 0.11
-CategoricalArrays 0.2.0
-Nulls
+CategoricalArrays 0.3.0
+Missings
 StatsBase 0.11.1
 Compat 0.9.2
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,5 @@
 julia 0.5
-DataTables
+DataFrames
 CategoricalArrays 0.0.6
 NullableArrays 0.0.10
 StatsBase 0.11.1
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
-julia 0.5
-DataFrames
-CategoricalArrays 0.0.6
-NullableArrays 0.0.10
+julia 0.6
+DataFrames 0.11
+CategoricalArrays 0.2.0
+Nulls
 StatsBase 0.11.1
 Compat 0.9.2
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ os:
   - linux
   - osx
 julia:
-  - 0.5
+  - 0.6
   - nightly
 notifications:
   email: false
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,14 +1,14 @@
 environment:
   matrix:
-  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.5/julia-0.5-latest-win32.exe"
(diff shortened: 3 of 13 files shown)


kleinschmidt:
cc @dmbates, I think you were asking for dataframes support here.

ararslan:
Interesting CI failures: It can't find `CategoricalArray` nor `Null`. Maybe CategoricalArrays and Nulls have to be added as test dependencies as well?

Also, can you add 0.6 to the CI testing while you're in here?

nalimilan:
No, that's because Nulls.jl isn't available on Julia 0.5. I've added Julia 0.6, should work there.

REQUIRE
```diff
@@ -1,6 +1,6 @@
-julia 0.5
-DataTables
-CategoricalArrays 0.0.6
-NullableArrays 0.0.10
+julia 0.6
+DataFrames 0.11
```
dmbates: There is no `DataFrames` `v0.11`.  How can this PR even be tested?

nalimilan: You need to checkout DataFrames from master. If `Pkg` complains about the version, you can call `include("test/runtests.jl")` directly. I guess I should have kept 0.10 for now.

dmbates: Actually you need to checkout the master version of `DataFrames` (and, in my experience, it is difficult to do this unless you first create a clean package directory) then clone `StatsModels` directly into the package directory without going through Julia's package system then check out the branch then run `runtests.jl` directly.

quinnj:
bump

nalimilan:
@quinnj What do you want to bump? This is blocked waiting for a DataFrames release to be tagged. Though I may have to do some adjustements to support Nulls 0.1.0.

quinnj:
I think it'd be worth merging this on master (and still waiting for a DataFrames tag to tag a new release here) to help in porting packages that have StatsModels as dependencies

nalimilan:
That won't really help, since dependencies will generally depend on DataFrames master too, isn't it? I'd rather tag a DataFrames release soon, which isn't going to be used by anybody anyway due to upper bounds on all dependencies.

quinnj:
> which isn't going to be used by anybody anyway due to upper bounds on all dependencies.

How so? If all I have in my package REQUIRES is `DataFrames` then as soon as we tag, it's going to upgrade and probably break that package I was using, right?

nalimilan:
No, with upper bounds the highest version compatible with all installed packages will be used. So basically, people will only use DataFrames once we have released compatible versions of all dependencies they use...

kleinschmidt:
Is anybody actually _using_ StatsModels? I think we need to have a clear upgrade path in place for people but I don't know how to go about that.  Maybe having a deprecation cycle with DataFrames where it depends on StatsModels and re-exports that functionality?

nalimilan:
Yeah, why not. In practice I don't see how to do that, though, since the exported functions from DataFrames would clash with those exported from StatsModels, wouldn't they?

nalimilan:
Good news: the tests pass when using DataArrays columns. I think this branch is ready to merge once DataFrames 0.11 has been tagged.

nalimilan:
I've added a commit to port to missings (tests pass locally).

Issue #32, opened 2017-09-11
Get rid of DataFrameStatisticalModel


nalimilan:
The current approach of wrapping fitted `StatisticalModel` objects within `DataFrameStatisticalModel`/`DataTableStatisticalModel` objects has a major limitation: we can only `@delegate` functions that we know about, i.e. only those defined in StatsBase. This sounds problematic in the long term, where custom model types defined in packages may want to support new operations.

A solution to this would be to move from the current composition approach (wrapper) to inheritance. `StatisticalModel` object would be required to have some predefined fields, like `coefnames`, `formula`, `modelmatrix` and `modelframe`. The `fit` method from StatsModels would set these fields after fitting the model, so that they can be accessed later both by generic methods and by custom methods defined for a given model type.

This solution would for example allow implementing a generic Wald test function in StatsBase (https://github.com/JuliaStats/StatsBase.jl/issues/300), which would use the coefficient names stored in the object. This is not possible with the current `DataFrameStatisticalModel` approach, where coefficient names can only be accessed by methods designed for `DataFrameStatisticalModel`, which therefore cannot live in StatsBase.

andreasnoack:
The key consideration here is how we generally provide a high and low level API's for regression models such that you can enjoy high level convenience while still have access to all the functionality without overhead when you need to fit millions of models. Hence, we'd name to make sure that fields with names are not baked into that whole setup. This was the original reason that `GLM`'s dependency on `DataFrames` was removed and `DataFrameStatisticalModel` introduced here.

nalimilan:
Yes, optional fields would remain `null`/`nothing` by default , e.g. if you don't want names for your coefficients.

andreasnoack:
I'm skeptical that this is right model. Parameter names are not needed for a low level API and they mainly make sense in connection with some kind of dataframe. In the future, we might also be able to fit models based on e.g. streaming data sources so having prespecified fields for all possible regression models seems overly restrictive. Hence, I'm in favor of something like the existing wrapper approach where high level convenience is added alongside the actual fitting procedures in a wrapper object.

nalimilan:
Why would coefficient names be specific to data frames? Couldn't you want coefficient names even when fitting directly from a matrix or from a streaming source?

[...]

PR #33, opened 2017-11-09
Updating syntax to Julia 0.7

Nosferican:


kleinschmidt:
There's a bunch of other `type`s, and probably other deprecated syntax too...maybe better to run the whole thing through femtocleaner?

Nosferican:
Excluding `statmodel.jl` which will probably get deleted #32 I found most was already up to date... Running it in nightly just gave me a `findfirst` which I could add to this pull as well.

kleinschmidt:
I'm seeing `type` in ModelFrame and Formula, and a bunch of the old parametric method syntax too (at least in contrasts.jl)

Nosferican:
I might have modified those locally beforehand. Will get another copy of Master to examine. Will make the edits after I have checked those.

Nosferican:
Ran femtocleaner (cool tool, thanks). Let me know if I can help with any of the edits or if those are good.

andreasnoack:
> Ran femtocleaner

There is no reason to do this by hand. The bot can do it automatically and open the necessary PR when it doesn't conflict with backward compatibility. So here we should just drop 0.5 support and it should happen automatically.

src/modelframe.jl
```diff
@@ -87,7 +87,7 @@ function check_non_redundancy!(trms::Terms, df::AbstractDataTable)
                 dropped = trms.factors[:,i_term]
                 dropped[i_eterm] = 0
 
-                if findfirst(encountered_columns, dropped) == 0
+                if findfirst(equalto(encountered_columns), dropped) == 0
```
nalimilan: This doesn't work on 0.6. Use `x -> x == encountered_columns` instead.

src/modelframe.jl
```diff
@@ -46,17 +46,17 @@ julia> mf = ModelFrame(y ~ 1 + x, df)
...
     terms::Terms
     msng::BitArray
     ## mapping from df keys to contrasts matrices
     contrasts::Dict{Symbol, ContrastsMatrix}
 end
 
-is_categorical{T<:Real}(::AbstractArray{T}) = false
-typealias NullableReal{T<:Real} Nullable{T}
-is_categorical{T<:NullableReal}(::AbstractArray{T}) = false
+is_categorical(::AbstractArray{T}) where {T<:Real} = false
```
nalimilan: Better use `AbstractArray{<:Real}`. Same below.

src/contrasts.jl
```diff
@@ -244,11 +244,11 @@ julia> StatsModels.ContrastsMatrix(StatsModels.FullDummyCoding(), ["a", "b", "c"
  0.0  0.0  0.0  1.0
 ```
 """
-type FullDummyCoding <: AbstractContrasts
+mutable struct FullDummyCoding <: AbstractContrasts
 # Dummy contrasts have no base level (since all levels produce a column)
 end
 
-ContrastsMatrix{T}(C::FullDummyCoding, levels::Vector{T}) =
+ContrastsMatrix(C::FullDummyCoding, levels::Vector{T}) where {T} =
```
nalimilan: `T` isn't used at all?

src/contrasts.jl
```diff
@@ -128,7 +128,7 @@ ContrastsMatrix(contrasts_matrix::ContrastsMatrix, levels::AbstractVector)
   constructing a model matrix from a `ModelFrame` using different data.
 
 """
-function ContrastsMatrix{C <: AbstractContrasts}(contrasts::C, levels::AbstractVector)
+function ContrastsMatrix(contrasts::C, levels::AbstractVector) where C <: AbstractContrasts
```
nalimilan: No need for `C` here nor below.

src/modelmatrix.jl
```diff
@@ -37,13 +37,13 @@ function modelmat_cols{T<:AbstractFloatMatrix}(::Type{T}, name::Symbol, mf::Mode
     end
 end
 
-modelmat_cols{T<:AbstractFloatMatrix, V<:AbstractRealVector}(::Type{T}, v::V) =
+modelmat_cols(::Type{T}, v::V) where {T<:AbstractFloatMatrix, V<:AbstractRealVector} =
```
nalimilan: `V::AbstractRealVector` would be enough here and with similar cases below.

src/contrasts.jl
kleinschmidt: I wrote this code when I was just getting handle on type parameters, so I went a little overboard (better safe than sorry?)

src/modelframe.jl
```diff
@@ -229,19 +229,19 @@ function coefnames(mf::ModelFrame)
     terms = droprandomeffects(dropresponse!(mf.terms))
 
     ## strategy mirrors ModelMatrx constructor:
-    eterm_names = @compat Dict{Tuple{Symbol,Bool}, Vector{Compat.UTF8String}}()
-    term_names = Vector{Compat.UTF8String}[]
+    eterm_names = Dict{Tuple{Symbol,Bool}, Vector{String}}()
+    term_names = Vector{String}[]
```
Nosferican: Should it be `Vector{String}()` instead?

kleinschmidt: Nope, it's a vector of vectors of strings :)

kleinschmidt: Could use `Vector{Vector{Strings}}()` which might be clearer.

Nosferican:
Thanks for the feedback. I will make these changes today and ping y'all. I believe there is a bug in `ModelMatrix` with categorical variables (`BoundsError: attempt to access`) so I want to get the depreciations out of the way and get the `DataFrames` / `DataTables` figure out so I can investigate the bug better.

src/modelframe.jl
Nosferican: I don't think there will be a compatible `DataFrames` / `StatsModels` version for Julia 0.6. Should the package still provide support for it? I believe 1.0 would be 0.7 without the depreciations mostly so it might be better to adjust it now.

nalimilan: Yes, there will be a 0.6 version AFAIK. Anyway the version I proposed will work everywhere, it's just slightly more verbose.

Nosferican:
Are we deleting StatsModels/statsmodel.jl or should I update it as well? #32

Nosferican:
I had never used the `{T<:}` syntax... Guess as a latecomer I learned from the start to use
```
function (obj::AbstractTypeOfobj) end
```
but I think I figured it out... regardless let me know after y'all revise it. Thanks.

src/contrasts.jl
```diff
@@ -79,25 +79,25 @@ termnames(C::MyCoding, levels, baseind) = ...
 ```
 
 """
-abstract AbstractContrasts
+abstract type AbstractContrasts end
 
 # Contrasts + Levels (usually from data) = ContrastsMatrix
-type ContrastsMatrix{C <: AbstractContrasts, T}
+mutable struct ContrastsMatrix{T}
```
nalimilan: This shouldn't be changed as it makes the type non-concrete. Only remove type parameters when they are used only once in a function's signature (if they are used twice, they enforce consistency of the particular concrete type which is passed). Same below in several places.

src/contrasts.jl
```diff
@@ -79,25 +79,25 @@ termnames(C::MyCoding, levels, baseind) = ...
...
 # will behave identically in creating modelmatrix columns
-Base.:(==){C<:AbstractContrasts,T}(a::ContrastsMatrix{C,T}, b::ContrastsMatrix{C,T}) =
+Base.:(==)(a::ContrastsMatrix{AbstractContrasts,T}, b::ContrastsMatrix{AbstractContrasts,T}) =
     a.matrix == b.matrix &&
     a.termnames == b.termnames &&
     a.levels == b.levels
 
-Base.hash{C}(a::ContrastsMatrix{C}, h::UInt) =
+Base.hash(a::ContrastsMatrix, h::UInt) =
     hash(C, hash(a.matrix, hash(a.termnames, hash(a.levels, h))))
```
nalimilan: `C` is used!

src/contrasts.jl
```diff
@@ -176,7 +176,7 @@ function ContrastsMatrix{C <: AbstractContrasts}(contrasts::C, levels::AbstractV
     ContrastsMatrix(mat, tnames, c_levels, contrasts)
 end
 
-ContrastsMatrix{C <: AbstractContrasts}(c::Type{C}, levels::AbstractVector) =
+ContrastsMatrix(c::AbstractContrasts, levels::AbstractVector) =
```
nalimilan: Should be `Type{<:AbstractContrasts}`.

src/modelframe.jl
```diff
@@ -107,7 +107,7 @@ const DEFAULT_CONTRASTS = DummyCoding
 _unique(x::CategoricalArray) = unique(x)
 _unique(x::NullableCategoricalArray) = [get(l) for l in unique(x) if !isnull(l)]
 
-function _unique{T<:Nullable}(x::AbstractArray{T})
+function _unique(x::AbstractArray{T<:Nullable})
```
nalimilan: Remove `T`.

src/modelmatrix.jl
```diff
@@ -26,7 +26,7 @@ Base.size(mm::ModelMatrix, dim...) = size(mm.m, dim...)
 
 
 ## construct model matrix columns from model frame + name (checks for contrasts)
-function modelmat_cols{T<:AbstractFloatMatrix}(::Type{T}, name::Symbol, mf::ModelFrame; non_redundant::Bool = false)
+function modelmat_cols(T::AbstractFloatMatrix, name::Symbol, mf::ModelFrame; non_redundant::Bool = false)
```
nalimilan: Should be `Type{<:AbstractFloatMatrix}`. Anyway `T` is used below so you need to keep it in this case.

src/modelmatrix.jl
```diff
@@ -37,47 +37,46 @@ function modelmat_cols{T<:AbstractFloatMatrix}(::Type{T}, name::Symbol, mf::Mode
     end
 end
 
-modelmat_cols{T<:AbstractFloatMatrix, V<:AbstractRealVector}(::Type{T}, v::V) =
-    convert(T, reshape(v, length(v), 1))
+modelmat_cols(T::AbstractFloatMatrix, v::AbstractRealVector) =
 # FIXME: this inefficient method should not be needed, cf. JuliaLang/julia#18264
-modelmat_cols{T<:AbstractFloatMatrix, V<:NullableRealVector}(::Type{T}, v::V) =
+modelmat_cols(T::AbstractFloatMatrix, v::NullableRealVector) =
```
nalimilan: Should be `Type{<:AbstractFloatMatrix}` here and repeatedly below.

src/modelmatrix.jl
```diff
@@ -37,47 +37,46 @@ function modelmat_cols{T<:AbstractFloatMatrix}(::Type{T}, name::Symbol, mf::Mode
     end
 end
 
-modelmat_cols{T<:AbstractFloatMatrix, V<:AbstractRealVector}(::Type{T}, v::V) =
-    convert(T, reshape(v, length(v), 1))
+modelmat_cols(T::AbstractFloatMatrix, v::AbstractRealVector) =
```
nalimilan: ?!

src/modelmatrix.jl
```diff
@@ -37,47 +37,46 @@ function modelmat_cols{T<:AbstractFloatMatrix}(::Type{T}, name::Symbol, mf::Mode
...
-                                               contrast::ContrastsMatrix)
+function modelmat_cols(T::AbstractFloatMatrix,
+    v::Union{CategoricalVector, NullableCategoricalVector},
+    contrast::ContrastsMatrix)
     ## make sure the levels of the contrast matrix and the categorical data
     ## are the same by constructing a re-indexing vector. Indexing into
     ## reindex with v.refs will give the corresponding row number of the
     ## contrast matrix
-    reindex = [findfirst(contrast.levels, l) for l in levels(v)]
+    reindex = [findfirst(equalto(contrast.levels), l) for l in levels(v)]
```
nalimilan: Remove `equalto` here too.

src/modelmatrix.jl
```diff
@@ -37,47 +37,46 @@ function modelmat_cols{T<:AbstractFloatMatrix}(::Type{T}, name::Symbol, mf::Mode
...
 end
 
-indexrows(m::SparseMatrixCSC, ind::Vector{Int}) = m'[:, ind]'
-indexrows(m::AbstractMatrix, ind::Vector{Int}) = m[ind, :]
+indexrows(m::SparseMatrixCSC, ind::AbstractVector{Int}) = m'[:, ind]'
+indexrows(m::AbstractMatrix, ind::AbstractVector{Int}) = m[ind, :]
 
 """
-    expandcols{T<:AbstractFloatMatrix}(trm::Vector{T})
+    expandcols(trm::AbstractVector{AbstractFloatMatrix})
```
nalimilan: Should be `<:AbstractFloatMatrix`, that's not the same thing.

nalimilan:
> Are we deleting StatsModels/statsmodel.jl or should I update it as well? #32

I don't think it's going to be removed soon, so better fix it too.

> I had never used the {T<:} syntax... Guess as a latecomer I learned from the start to use
> 
> function (obj::AbstractTypeOfobj) end
> 
> but I think I figured it out... regardless let me know after y'all revise it. Thanks.

No, that's not the same thing. That syntax has always existed, `{T <: U}` was just an equivalent of the new `where {T <: U}` syntax. BTW, if you had ran the test locally you'd have found out cases where the replacements you made are not equivalent.

src/contrasts.jl
```diff
@@ -79,25 +79,25 @@ termnames(C::MyCoding, levels, baseind) = ...
...
+abstract type AbstractContrasts end
 
 # Contrasts + Levels (usually from data) = ContrastsMatrix
-type ContrastsMatrix{C <: AbstractContrasts, T}
+mutable struct ContrastsMatrix{C <: AbstractContrasts, T}
     matrix::Matrix{Float64}
     termnames::Vector{T}
     levels::Vector{T}
-    contrasts::C
+    contrasts::AbstractContrasts
```
nalimilan: Still need to use `C` here.

src/contrasts.jl
```diff
@@ -79,25 +79,25 @@ termnames(C::MyCoding, levels, baseind) = ...
...
 end
 
 # only check equality of matrix, termnames, and levels, and that the type is the
 # same for the contrasts (values are irrelevant).  This ensures that the two
 # will behave identically in creating modelmatrix columns
-Base.:(==){C<:AbstractContrasts,T}(a::ContrastsMatrix{C,T}, b::ContrastsMatrix{C,T}) =
+Base.:(==)(a::ContrastsMatrix{AbstractContrasts,T}, b::ContrastsMatrix{AbstractContrasts,T}) =
```
nalimilan: Same here.

src/modelmatrix.jl
```diff
@@ -37,47 +37,47 @@ function modelmat_cols{T<:AbstractFloatMatrix}(::Type{T}, name::Symbol, mf::Mode
     end
 end
 
-modelmat_cols{T<:AbstractFloatMatrix, V<:AbstractRealVector}(::Type{T}, v::V) =
-    convert(T, reshape(v, length(v), 1))
+modelmat_cols(::Type{T}, v::V) where {T <: AbstractFloatMatrix, V<:AbstractRealVector} =
+convert(T, reshape(v, length(v), 1))
```
nalimilan: Keep indentation.

src/modelmatrix.jl
```diff
@@ -37,47 +37,47 @@ function modelmat_cols{T<:AbstractFloatMatrix}(::Type{T}, name::Symbol, mf::Mode
...
 # FIXME: this inefficient method should not be needed, cf. JuliaLang/julia#18264
-modelmat_cols{T<:AbstractFloatMatrix, V<:NullableRealVector}(::Type{T}, v::V) =
+modelmat_cols(::Type{<:AbstractFloatMatrix}, v::V) where {T <: AbstractFloatMatrix, V <: NullableRealVector} =
```
nalimilan: Should be `::Type{T}`. Same below (multiple times).

src/modelmatrix.jl
```diff
@@ -65,7 +65,7 @@ function modelmat_cols(::Type{T},
     ## are the same by constructing a re-indexing vector. Indexing into
     ## reindex with v.refs will give the corresponding row number of the
     ## contrast matrix
-    reindex = [findfirst(x -> x == contrast.levels, l) for l in levels(v)]
+    reindex = [findfirst(x -> x == l, contrast.levels) for l in levels(v)]
```
ararslan: For an appropriate Compat version, this can be `findfirst(equalto(l), contrast.levels)`

nalimilan: Yeah, why not, though the current version isn't terrible either.

Nosferican:
I was only able to get the module running in 0.7 after commenting out the `statsmodels.jl`... also had to address some minor issues with DataTables -> DataFrames and the `NullableArrays`... However, I am waiting for some of those changes to further test `contrasts.jl` which was giving me a bound error with `CategoricalArrays`.

src/modelmatrix.jl
ararslan: The only difference is that `equalto` uses `isequal`. Probably shouldn't matter in this case, unless any of the levels are `NaN` or `missing` or some other irregular thing.

Nosferican:
`!(x in group)` as efficient as `x ∉ group` or is there any advantage to using the latter?

ararslan:
They're the same

Nosferican:
My mistake, I pushed something to my master before the merge. Correcting that.

REQUIRE
```diff
@@ -1,6 +1,5 @@
-julia 0.6
-DataTables
-CategoricalArrays 0.0.6
-NullableArrays 0.0.10
-StatsBase 0.11.1
-Compat 0.9.2
+julia 0.7
+DataFrames 0.10.1+
```
ararslan: These shouldn't have a trailing +

appveyor.yml
```diff
@@ -1,14 +1,14 @@
...
+  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.5/julia-0.6-latest-win32.exe"
+  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.5/julia-0.6-latest-win64.exe"
   - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x86/julia-latest-win32.exe"
   - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x64/julia-latest-win64.exe"
 
 branches:
   only:
     - master
-    - /release-.*/
+    - release-0.6
```
ararslan: This change is incorrect; this refers to the branches that trigger AppVeyor builds rather than release versions of Julia. For example, a pull request against a branch called release-0.11 on this repository should trigger a build but with this change it would not.

nalimilan:
I've just merged my PR porting to DataFrames 0.11, which automatically triggered Femtocleaner. Sorry for all the conflicts it introduced, but it should also have fixed most deprecations. Not sure whether some still need to be fixed manually.

src/modelframe.jl
```diff
@@ -239,7 +240,8 @@ function coefnames(mf::ModelFrame)
     for (i_term, term) in enumerate(terms.terms)
 
         ## names for columns for eval terms
-        names = Vector{String}[]
+
```
nalimilan: Remove this.

src/contrasts.jl
```diff
@@ -114,7 +114,7 @@ for the base level (which defaults to the first level).
 # Constructors
 
 ```julia
-ContrastsMatrix{C <: AbstractContrasts}(contrasts::C, levels::AbstractVector)
```
nalimilan: I think the parameter should be kept (else it's a duplicate of the following line).

src/modelframe.jl
Nosferican: Change back `Vector{Vector{String}}()` for `Vector{String}[]`?

nalimilan: Sorry, I meant the new empty line. Same at line 144.

src/contrasts.jl
nalimilan: "Kept" doesn't mean "remove the line". That parameter was legitimate AFAICT if you want to specify `C`.

Nosferican: The line is not part of the code. It is part of the documentation for constructors. I verified if any of the constructors used `C` and none of them did. Since none of the constructors methods use `C` it seems misleading to document that method when in fact it is equivalent in all regards to the other.

Nosferican:
I reordered the lines in Require in order of dependencies because I had issues which would not compile the module when they were in certain orders (for DataFrames and StatsModels). I don't know if that is an issue for Julia 0.7 or something unrelated, but the order did seem to make it more robust.

test/contrasts.jl
```diff
@@ -42,7 +44,7 @@ end
                             1  0  0
                             1  0  0
                             1  1  0]
-@test coefnames(mf) == ["(Intercept)"; "x: b"; "x: c"]
+@test StatsBase.coefnames(mf) == ["(Intercept)"; "x: b"; "x: c"]
```
nalimilan: No need to fully qualify exported functions when calling them (only when overloading them).

Nosferican: `coefnames` is not exported in the latest tagged StatsBase (it is in master, but not in the latest release which is the one Travis uses). As soon as the next StatsBase version is tagged then we can drop the prefix.

nalimilan: Then using the fully qualified name hides a breaking change: `coefnames` used to be exported by `StatsModels`, and now it's no longer available. Unfortunately, I'm not sure there's a good way to handle this: deprecating `coefnames` here but still exporting it creates a name clash with StatsBase even if that's the same function. I guess we could remove `coefnames` without prior notice, knowing that StatsModels is very young and everything is in transition yet.

At any rate, better require StatsBase 0.19.2: https://github.com/JuliaLang/METADATA.jl/pull/12279

Nosferican: Thanks for the tag. Since StatsModels is meant to be used with `StatsBase` either with `StatsModels.DataFrameStatisticalModel` or `StatsBase.StatisticalModel` should we `@rexport` StatsBase and DataFrames?
It makes more sense for
```
using StatsModels
```
rather than
```
using DataFrames, StatsBase, StatsModels
```

nalimilan: Maybe, but that's a difficult question that should be tackled separately.

I've merged the StatsBase 0.19.2 PR.

Nosferican:
Should be ready to merge now.

src/formula.jl
```diff
@@ -109,7 +109,7 @@ function distribute!(ex::Expr)
             typeof(e)==Expr && e.head == :call && e.args[1] == distributing_op
 
         ## find first distributing subex
-        first_distributing_subex = findfirst(is_distributing_subex, ex.args)
+        first_distributing_subex = findfirst(equalto(ex.args), is_distributing_subex)
```
kleinschmidt: `is_distributing_subex` is a predicate function so I think the original is correct.

Nosferican: It might be the case, I am running the tests locally now to verify.

Nosferican:
Thanks to #42 now it has an all clear in 0.6 and nightly. I believe I got all depreciations now, but feel free to one last check just in case. If all good, it should be good to merge. Thanks y'all for all the feedback!

nalimilan:
Thanks!

PR #42, opened 2017-11-28
Change delegate macro to avoid a QuoteNode

### LINKED ISSUE #41: delegate macro fails to precompile in v0.7.0-dev
@andreasnoack I hate to bother you on this but `git blame` indicates that you were the last person to fix the `delegate` macro in `src/statsmodel.jl`.  Right now precompilation of the `DataFrames11` branch fails under v0.7.0-dev with
```julia
julia> using StatsModels  
INFO: Precompiling module StatsModels.              
ERROR: LoadError: LoadError: LoadError: type QuoteNode has no field args                                 
Stacktrace:               
 [1] @delegate(::LineNumberNode, ::Module, ::Any, ::Any) at /home/bates/.julia/v0.7/StatsModels/src/statsmodel.jl:20
 [2] include_relative(::Module, ::String) at ./loading.jl:509                                            
 [3] include at ./sysimg.jl:15 [inlined]            
 [4] include(::String) at /home/bates/.julia/v0.7/StatsModels/src/StatsModels.jl:3                       
 [5] collect_to!(::Array{Any,1}, ::Base.Generator{Array{String,1},typeof(StatsModels.include)}, ::Int64, ::Int64) at ./array.jl:603
 [6] collect_to!(::Array{Function,1}, ::Base.Generator{Array{String,1},typeof(StatsModels.include)}, ::Int64, ::Int64) at ./array.jl:613 (repeats 2 times)
 [7] collect_to_with_first!(::Array{typeof(StatsModels.contrasts_matrix),1}, ::Function, ::Base.Generator{Array{String,1},typeof(StatsModels.include)}, ::Int64) at ./array.jl:590
 [8] _collect(::Array{String,1}, ::Base.Generator{Array{String,1},typeof(StatsModels.include)}, ::Base.EltypeUnknown, ::Base.HasShape) at ./array.jl:584
 [9] collect_similar(::Array{String,1}, ::Base.Generator{Array{String,1},typeof(StatsModels.include)}) at ./array.jl:520
 [10] map(::Function, ::Array{String,1}) at ./abstractarray.jl:1901                                      
 [11] top-level scope     
 [12] include_relative(::Module, ::String) at ./loading.jl:509                                           
 [13] include(::Module, ::String) at ./sysimg.jl:15 
 [14] top-level scope     
 [15] top-level scope at ./<missing>:2              
in expression starting at /home/bates/.julia/v0.7/StatsModels/src/statsmodel.jl:62                       
in expression starting at /home/bates/.julia/v0.7/StatsModels/src/statsmodel.jl:62                       
in expression starting at /home/bates/.julia/v0.7/StatsModels/src/StatsModels.jl:26                      
ERROR: Failed to precompile StatsModels to /home/bates/.julia/lib/v0.7/StatsModels.ji.                   
Stacktrace:               
 [1] compilecache(::String) at ./loading.jl:637     
 [2] _require(::Symbol) at ./loading.jl:448         
 [3] require(::Symbol) at ./loading.jl:300          
```

To me this is an indication that these exotic model abstract types might do with some reconsideration but, for the short term, it would help if that could be repaired.

andreasnoack:
Should fix #41 although I'm a bit surprised that this makes it an `Expr` instead of `QuoteNode`. This will probably give a test error on master because of https://github.com/JuliaLang/julia/issues/24825


andreasnoack:
Hm. Something broke on 0.6 but I ended up in a `Pkg` deadlock on 0.6 because of the upper bounds introduced on many packages and I'll have to stop for tonight. This deadlocking seems to be a pretty common consequence of adding upper bounds.

Nosferican:
I would actually prefer to have the breaking change from `StatsModels.DataFrameStatisticalModel` to `StatsBase.StatisticalModel` inheritance for the first working release to Julia 0.7.

nalimilan:
Failure on 0.6 seems legitimate:
```
  Expression: model_response(m) == Array(d[:y])
  TypeError: getfield: expected Symbol, got Expr
```

andreasnoack:
Indeed, the problem is that my local system is still completely deadlocked because of the upper bounds.

andreasnoack:
Now it works. As a consequence of the discussion in https://github.com/JuliaLang/julia/issues/24825, I decided to change the macro to return an expression with an error instead of throwing from the macro. It seems that throwing from macros will give confusing behavior (non-interceptable exceptions that behave differently when `eval(parse(...))`ed) and error messages (interpolated `LoadError`s) so delaying the error to runtime is probably better. Thoughts?

nalimilan:
That makes sense. Unless you can see any drawbacks to that approach, I'd say it's fine.

PR #44, opened 2017-11-30
qualify coefnames when extending

kleinschmidt:
Necessary to prevent clash with StatsModels


kleinschmidt:
dmbates/MixedModels.jl/#105

PR #45, opened 2017-11-30
Changes for v0.11 of DataFrames

dmbates:
Also some code cleanup.  `Compat.view` can now be `view`.  In the tests `Base.Test` -> `Compat.Test`

PR #46, opened 2017-12-04
Fix warnings from `Pkg.test` on v0.6.1

dmbates:
I extracted these two simple fixes from #28 and #45 which were more complicated and, to some extent, overridden by #33.  They eliminate warnings from `Pkg.test("StatsModels)` in v0.6.1 of Julia.


dmbates:
Travis failures on nightlies are build errors for `SpecialFunctions`.

Issue #31, opened 2017-09-02
default intercept behavior


kleinschmidt:
I'd like to change the default intercept behavior in formulas.  Currently, we have this:

* `~ 1 + x` intercept column and x
* `~ x` intercept column and x
* `~ 0 + x` or `~ -1 + x` or `~ x-1` no intercept column, only x.

Including the intercept by default might be sensible when all you're doing is regression. But it's not always the most appropriate thing (e.g. ararslan/Survival.jl#2) and making people use `~ 0+x` when they don't want an intercept feels like an unnecessary gotcha, Especially if we want formulas to be useful as a general "glue" between tabular data and numerical arrays.

I think we've largely agreed that this is the desirable behavior:

* `~ 1+x` intercept and x
* `~ x` just x
* `~ 0+x` or `~ -1+x` or `~ x-1` an (informative) error.

If that's the case, let's implement that before we release this as a replacement for the statsmodels code in DataFrames.

piever:
Just to repeat here some concern that I was expressing in [ararslan/Survival.jl#2](https://github.com/ararslan/Survival.jl/pull/2#issuecomment-322272456), for Cox regression we need a formula such that, if we only have a variable `x`:
- if `x` is continuous, then the number of columns in regression matrix is equal to one
- if `x` is categorical and can take `n` values, then the number of columns in regression matrix is equal to `n-1`

Would there be a syntax for that?

ararslan:
Your first bullet point should be the default if there's no intercept term. That would just be `~ x`. The second bullet point is interesting... I kind of think that should be the default for categorical vectors, but there has to be a way to specify the reference level.

piever:
I'd also tend to think that `~ 1+x` should have exactly one column more than `~ x` or at the very least find it puzzling that the two things should be different for continuous variables but equal/equivalent for categorical. Once we take `~ x` to be the same as `~ 1+x` in the categorical case, it becomes a bit hard to express that we want one column less.

nalimilan:
I think the Cox model issue is mostly orthogonal to this issue: whether `~ x` includes an implicit intercept or not when `x` is continuous, we will need a way to indicate that the model matrix shouldn't be full rank when `x` is categorical. So we need another mechanism anyway.

[...]

PR #47, opened 2017-12-12
added has_intercept

piever:
I've added the option of manually specifying that a specific model doesn't want the intercept column (application case: Cox model, where the intercept column is meaningless and should be excluded, see discussion in [#2](https://github.com/ararslan/Survival.jl/pull/2) ). This happens with a keyword `intercept = true` to `ModelFrame`. When building a ModelFrame to fit a model from data, this keyword will default to `has_intercept(T)` where `T` is your model type. To add the desired behaviour for your model (in my case CoxModel) you can simply overload it in a specific package. For example, in my case it would be:

```julia
StatsModels.has_intercept(::Type{CoxModel}) = false
```

I know there has been some discussion as to what is the default behaviour for intercept (add it by default or not, see #31 ), but here I'd actually prefer to sidestep that discussion and only add the possibility to exclude the intercept for some models.

### DIFF
--- a/src/modelframe.jl
+++ b/src/modelframe.jl
@@ -140,7 +140,9 @@ _droplevels!(x::Any) = x
 _droplevels!(x::AbstractCategoricalArray) = droplevels!(x)
 
 function ModelFrame(trms::Terms, d::AbstractDataFrame;
-                    contrasts::Dict = Dict())
+                    contrasts::Dict = Dict(), intercept::Bool = true)
+    trms = Terms((getfield(trms, fieldname) for fieldname in fieldnames(trms))...)
+    intercept || (trms.intercept = false)
     df, msng = missing_omit(DataFrame(map(x -> d[x], trms.eterms)))
     names!(df, Symbol.(string.(trms.eterms)))
 
--- a/src/modelframe.jl
+++ b/src/modelframe.jl
@@ -140,9 +140,7 @@ _droplevels!(x::Any) = x
 _droplevels!(x::AbstractCategoricalArray) = droplevels!(x)
 
 function ModelFrame(trms::Terms, d::AbstractDataFrame;
-                    contrasts::Dict = Dict(), intercept::Bool = true)
-    trms = Terms((getfield(trms, fieldname) for fieldname in fieldnames(trms))...)
-    intercept || (trms.intercept = false)
+                    contrasts::Dict = Dict())
     df, msng = missing_omit(DataFrame(map(x -> d[x], trms.eterms)))
     names!(df, Symbol.(string.(trms.eterms)))
 
--- a/src/statsmodel.jl
+++ b/src/statsmodel.jl
@@ -46,13 +46,16 @@ struct DataFrameRegressionModel{M,T} <: RegressionModel
 end
 
 """
-    supports_intercept(::Any)
+    supports_intercept(::Type)
 
-Defines whether a given model supports intercept. Returns `true`. To specify that a model `T` does not support an intercept, overload this function for the corresponding type: `supports_intercept(::Type{T}) = false`
+Define whether a given model supports intercept. Return `true` by default. To specify 
+that a model type `T` does not support an intercept, overload this function for the 
+corresponding type: `supports_intercept(::Type{T}) = false`
 
-Models that do not support intercept will be fit without one: the intercept term will be removed even if explicitly provided by the user.
+Models that do not support an intercept will be fitted without one: the intercept term 
+will be removed even if explicitly provided by the user.
 """
-supports_intercept(::Any) = true
+supports_intercept(::Type) = true
 
 for (modeltype, dfmodeltype) in ((:StatisticalModel, DataFrameStatisticalModel),
                                  (:RegressionModel, DataFrameRegressionModel))
--- a/src/statsmodel.jl
+++ b/src/statsmodel.jl
@@ -45,15 +45,23 @@ struct DataFrameRegressionModel{M,T} <: RegressionModel
     mm::ModelMatrix{T}
 end
 
-# Overload this function for a specific model type, according to whether it requires an intercept
-has_intercept(::Any) = true
+"""
+    supports_intercept(::Any)
+
+Defines whether a given model supports intercept. Returns `true`. To specify that a model `T` does not support an intercept, overload this function for the corresponding type: `supports_intercept(::Type{T}) = false`
+
+Models that do not support intercept will be fit without one: the intercept term will be removed even if explicitly provided by the user.
+"""
+supports_intercept(::Any) = true
 
 for (modeltype, dfmodeltype) in ((:StatisticalModel, DataFrameStatisticalModel),
                                  (:RegressionModel, DataFrameRegressionModel))
     @eval begin
         function StatsBase.fit(::Type{T}, f::Formula, df::AbstractDataFrame,
                                args...; contrasts::Dict = Dict(), kwargs...) where T<:$modeltype
-            mf = ModelFrame(f, df, contrasts=contrasts, intercept=has_intercept(T))
+            trms = Terms(f)
+            supports_intercept(T) || (trms.intercept = false)
+            mf = ModelFrame(trms, df, contrasts=contrasts)
             mm = ModelMatrix(mf)
             y = model_response(mf)
             $dfmodeltype(fit(T, mm.m, y, args...; kwargs...), mf, mm)
--- a/src/statsmodel.jl
+++ b/src/statsmodel.jl
@@ -46,16 +46,17 @@ struct DataFrameRegressionModel{M,T} <: RegressionModel
 end
 
 """
-    supports_intercept(::Type)
+    drop_intercept(::Type)
 
-Define whether a given model supports intercept. Return `true` by default. To specify 
-that a model type `T` does not support an intercept, overload this function for the 
-corresponding type: `supports_intercept(::Type{T}) = false`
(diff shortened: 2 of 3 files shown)


src/modelframe.jl
```diff
@@ -140,7 +140,9 @@ _droplevels!(x::Any) = x
 _droplevels!(x::AbstractCategoricalArray) = droplevels!(x)
 
 function ModelFrame(trms::Terms, d::AbstractDataFrame;
-                    contrasts::Dict = Dict())
+                    contrasts::Dict = Dict(), intercept::Bool = true)
+    trms = Terms((getfield(trms, fieldname) for fieldname in fieldnames(trms))...)
```
piever: I thought it was better to make a copy, in case some user wants to reuse these terms for a different model, but maybe there's no need for this line

nalimilan: Makes sense, but isn't there a simpler way to make a copy of `Terms` objects, like `copy`?

src/statsmodel.jl
```diff
@@ -45,12 +45,15 @@ struct DataFrameRegressionModel{M,T} <: RegressionModel
     mm::ModelMatrix{T}
 end
 
+# Overload this function for a specific model type, according to whether it requires an intercept
```
nalimilan: Can you turn this comment into a docstring, and develop it a little bit more (e.g. what function uses it, and what does it do exactly when it is set)?

src/modelframe.jl
piever: Surprisingly not... With `copy` I got:

```julia
julia> copy(StatsModels.Terms(@formula(x~y)))
ERROR: MethodError: no method matching copy(::StatsModels.Terms)
```

nalimilan: OK. Probably worth defining it then. It's quite easy: just call the constructor and pass it all the fields (arrays do not need to be copied, that's `deepcopy`'s job).

src/statsmodel.jl
piever: Sure. I'm also not 100% sure about the name. Now it defaults to `true` and only is effective when set to `false`, in which case it removes the intercept, so maybe it'd be clearer to instead define `exclude_intercept` defaulting to `false`. What do you think?

nalimilan: "allows" sounds like it would throw an error if it was requested. Maybe `include_intercept` or `use_intercept`? Or `supports_intercept`, which refers more directly to the model type?

piever: I like `supports_intercept`. Should the `ModelFrame` keyword be the same? I.e. `ModelFrame(f, df, contrasts=contrasts, supports_intercept=supports_intercept(T))`

nalimilan: OK. I'd keep `intercept::Bool` for `ModelFrame`, as you really decide whether you want an intercept or not in that case (it could even be used for models which support intercepts).

piever: Another option would be to not add any keyword to ModelFrame and simply pass it the modified `Terms` (already without intercept)

piever: If we keep an `intercept` keyword, with an eye on all models, I'd be in favor of not restricting to Bool but allow:
- `intercept = :auto` -> For now, doesn't do anything, in the future it will resorts to the defaults for the given Model (unless explicitly provided by the use) and error if the user didn't provide anything 
- `intercept = true` -> Adds the intercept
- `intercept = false` -> Removes the intercept

As it is in this PR the asymmetry may be confusing (`true` doesn't do anything and `false` removes the intercept). Right now `ModelFrame(@formula(y ~x), df, intercept = true)` would actually have no intercept.

nalimilan: A boolean would make sense if we call it `supports_intercept`: then the asymmetry would be OK. But it's weird to have this argument to `ModelFrame`.

Modifying the `Terms` object sounds simpler. I guess @kleinschmidt will have an opinion, but AFAICT better apply the smallest possible change, at least for now.

piever: Sure, I can update the PR as soon as a consensus is reached. The `Terms` object approach seems the simplest and most sensible to me (and the one with the smallest diff).

src/modelframe.jl
ararslan: Comment regarding the code rather than the content: `fieldnames` takes a type on 0.7, so we should make sure we have Compat with an appropriate lower bound as a dependency.

piever:
Following the discussion with @nalimilan, we figured out that actually it's much simpler to not add a keyword to `ModelFrame` but simply edit the terms that are passed to it by the `fit` function (`Terms` are obtained from the formula and changing them doesn't affect the formula, so there's also no need to copy them). With these changes the PR is actually very small. The function has been renamed `supports_intercept` and indicated whether it makes sense or not for a given statistical model to have an intercept column. Added docstrings.

src/statsmodel.jl
```diff
@@ -45,12 +45,23 @@ struct DataFrameRegressionModel{M,T} <: RegressionModel
     mm::ModelMatrix{T}
 end
 
+"""
+    supports_intercept(::Any)
```
nalimilan: Would better be `::Type` IIUC, right?

Also wrap lines at 92 chars, and use imperative ("define") rather than third person. Better say "Return `true` by default.", and "model type" rather than "model". Also "will be fitted".

nalimilan:
Ah, and could you add a test of this feature to test/statsmodel.jl? You can just copy the dummy model type to create your custom type that doesn't support intercepts.

piever:
Sure, I've added a test to make sure that, regardless of what the user gives, there is no intercept in the final result. I agree that maybe there will be future work to improve the whole intercept situation but for now I just need enough to be able to continue the work on the Survival package.

nalimilan:
> Could you also add tests to make sure this works how you'd expect/want when there are categorical variables involved?

@piever To expand on this, it turns out that more changes are probably required to handle categorical variables. Indeed, if I'm not mistaken, we need a categorical variable with `N` levels to be turned into `N-1` dummies, not `N`, even if there's no intercept. Apparently that's not how it works currently according to @kleinschmidt.

piever:
This is going to sound absurd, but actually this allowed Cox tests to pass because of #48 : with a dependent variable of type EventTime, I get N-1 dummies (whereas with a Int dependent variable I get N-1). I'll try to add `is_categorical_full_rank` (similarly to `supports_intercept`): in the meanwhile it would be good to figure out this EventTime anomaly...

nalimilan:
Funny.

Is there a real use case for `is_categorical_full_rank` being separate from `supports_intercept`? It seems to me that if the model family doesn't support an intercept, it doesn't make sense to expand categorical variables to full rank. Do we know of model families which would need that?

piever:
We commented the same thing at the same time, I think it's evidence enough to only have `supports_intercept` :)

nalimilan:
OK, so let's go with the simpler version, we can always extend it if we find a use case.

kleinschmidt:
I wonder if calling this something like `implicit_intercept` might make the behavior clearer: you want the model matrix to behave as if there is an intercept (for the purposes of rank checking) but never actually include one explicitly.

piever:
It's true, in this sense the simplest implementation is actually to add the intercept to the formula, compute the matrix and then remove the intercept column...

nalimilan:
Or maybe `drop_intercept`? `implicit_intercept` makes it sound like it affects whether an intercept term is added when you omit it in the formula, which is a totally different question.

piever:
`drop_intercept` is probably a better name, but @kleinschmidt idea of implementation works very well: I simply need to add the intercept, compute the ModelFrame and then remove the intercept. This should work for all cases, I'll update the tests and docstrings and push the new version in a bit.

piever:
Updated tests and docs, this version makes Cox tests pass  without using the #48 bug... I've added a test for the categorical case.

src/statsmodel.jl
```diff
@@ -45,12 +45,28 @@ struct DataFrameRegressionModel{M,T} <: RegressionModel
...
 for (modeltype, dfmodeltype) in ((:StatisticalModel, DataFrameStatisticalModel),
                                  (:RegressionModel, DataFrameRegressionModel))
     @eval begin
         function StatsBase.fit(::Type{T}, f::Formula, df::AbstractDataFrame,
                                args...; contrasts::Dict = Dict(), kwargs...) where T<:$modeltype
-            mf = ModelFrame(f, df, contrasts=contrasts)
+            trms = Terms(f)
+            drop_intercept(T) && (trms.intercept = true)
+            mf = ModelFrame(trms, df, contrasts=contrasts)
+            drop_intercept(T) && (mf.terms.intercept = false)
```
nalimilan: Why do you need to call this again? If that's needed, would it be possible to group the two `drop_intercept(T)` calls in a single `if` block?

piever: I need the first call to  make sure that the categorical variable is expanded in the rank-reduced form (N-1 levels) and the second call to remove the intercept column. Otherwise this would fail if the user gave a categorical variable without intercept (I would get one level too many). There needs to be the ModelFrame call in between so I can't really make a single if block.

nalimilan: OK, looks a bit weird, but well...

kleinschmidt: you'll either duplicate the modelframe call or the drop_intercept call...

nalimilan: Yeah, duplicating the `drop_intercept` call is better since it will be resolved statically.

nalimilan:
I've tagged a release: https://github.com/JuliaLang/METADATA.jl/pull/12491

Issue #49, opened 2018-01-12
Need to de-unionize model_response


kleinschmidt:
When the response (left-hand side) variable is a `Vector{Union{Missing,T}}`, the result of `model_response` is also a union type.  We need to change that method to convert to "de-unionize".  For RHS variables, this is handled by `modelmat_cols` methods that explicitly convert to a `AbstractFloatMatrix`, but I'm not sure it's always appropriate to convert to a vector of floats since there may be cases where having a non-float (categorical) response actually makes sense.

Related: JuliaStats/GLM.jl#210

kleinschmidt:
closed via #50

PR #50, opened 2018-01-12
de-unioninze model_response

kleinschmidt:
Addresses #49 (and possibly JuliaStats/GLM.jl#210).

Before `Missings`, all that was needed to ensure the `model_response` was a standard vector was to convert it to an `Array` (from a `DataArray`).  Now, with missing values potentially represented in a `Array{Union{T,Missing}}`, it's also necessary to use something like `Missings.disallowmissing`.  This won't work with missings in Base (in 0.7), but seeing as we use other features like `Missings.T` elsewhere (like uniqueifying categorical vectors, https://github.com/JuliaStats/StatsModels.jl/blob/d7f285a9f6c8a98943391c9995bd1b3c3614e8a5/src/modelframe.jl#L105-L115) I think it's good enough for now.

The future-proof alternative would be to define those methods here since I don't think they're planned to be supported in Base, or wait until we have something like `typesubtract`.

### DIFF
--- a/src/modelframe.jl
+++ b/src/modelframe.jl
@@ -177,7 +177,7 @@ Extract the response column, if present.  `DataVector` or
 """
 function StatsBase.model_response(mf::ModelFrame)
     if mf.terms.response
-        convert(Array, mf.df[mf.terms.eterms[1]])
+        Missings.disallowmissing(convert(Array, mf.df[mf.terms.eterms[1]]))
     else
         error("Model formula one-sided")
     end
--- a/src/modelframe.jl
+++ b/src/modelframe.jl
@@ -177,7 +177,8 @@ Extract the response column, if present.  `DataVector` or
 """
 function StatsBase.model_response(mf::ModelFrame)
     if mf.terms.response
-        Missings.disallowmissing(convert(Array, mf.df[mf.terms.eterms[1]]))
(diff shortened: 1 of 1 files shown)


src/modelframe.jl
```diff
@@ -177,7 +177,7 @@ Extract the response column, if present.  `DataVector` or
 """
 function StatsBase.model_response(mf::ModelFrame)
     if mf.terms.response
-        convert(Array, mf.df[mf.terms.eterms[1]])
+        Missings.disallowmissing(convert(Array, mf.df[mf.terms.eterms[1]]))
```
nalimilan: Since there's already a call to `convert` (and provided that it cannot be removed, i.e. that we really need an `Array`), better do `convert(Array{Missings.T(eltype(mf.df[mf.terms.eterms[1]])}, mf.df[mf.terms.eterms[1]])` (or something prettier using a temporary variable). That will avoid making an intermediate copy in some cases.

nalimilan:
Looks good in general (see my comment on implementation). But why do you say this won't work on 0.7? Missings.jl just reexports `Base.Missing` there.

quinnj:
Yeah, the `Missings.jl` package will continue to exist for the foreseeable future, for compat but also for other methods like `Missings.T` that need to get totally ironed out.

PR #51, opened 2018-01-15
Fix small issue in deunionization fix

andreasnoack:
...and remove a deprecation warning.


nalimilan:
Looks like we need a new test to cover this?

andreasnoack:
> Looks like we need a new test to cover this?

Indeed. Actually, the test was there but disabled. I've enabled it (and removed an obsolete test)

kleinschmidt:
Oof, sorry, that's what I get for comitting untested code.

PR #53, opened 2018-04-13
WIP updating for 0.7

kleinschmidt:
Lots of deprecation warnings still, and some of my fixes are pretty hacky, but tests pass locally.

### DIFF
--- a/src/contrasts.jl
+++ b/src/contrasts.jl
@@ -137,7 +137,7 @@ function ContrastsMatrix(contrasts::AbstractContrasts, levels::AbstractVector)
     #    better to filter data frame first
     # 3. contrast levels missing from data: would have empty columns, generate a
     #    rank-deficient model matrix.
-    c_levels = get(contrasts.levels, levels)
+    c_levels = contrasts.levels === nothing ? levels : contrasts.levels
     if eltype(c_levels) != eltype(levels)
         throw(ArgumentError("mismatching levels types: got $(eltype(levels)), expected " *
                             "$(eltype(c_levels)) based on contrasts levels."))
@@ -160,11 +160,11 @@ function ContrastsMatrix(contrasts::AbstractContrasts, levels::AbstractVector)
     end
 
     # find index of base level. use contrasts.base, then default (1).
-    baseind = isnull(contrasts.base) ?
+    baseind = contrasts.base === nothing ?
               1 :
-              Compat.findfirst(isequal(get(contrasts.base)), c_levels)
+              Compat.findfirst(isequal(contrasts.base), c_levels)
     if baseind === nothing
-        throw(ArgumentError("base level $(get(contrasts.base)) not found in levels " *
+        throw(ArgumentError("base level $(contrasts.base) not found in levels " *
                             "$c_levels."))
     end
 
@@ -199,24 +199,17 @@ function termnames(C::AbstractContrasts, levels::AbstractVector, baseind::Intege
     levels[not_base]
 end
 
-nullify(x::Nullable) = x
-nullify(x) = Nullable(x)
-
 # Making a contrast type T only requires that there be a method for
 # contrasts_matrix(T,  baseind, n) and optionally termnames(T, levels, baseind)
 # The rest is boilerplate.
 for contrastType in [:DummyCoding, :EffectsCoding, :HelmertCoding]
     @eval begin
         mutable struct $contrastType <: AbstractContrasts
-            base::Nullable{Any}
-            levels::Nullable{Vector}
+            base::Any
+            levels::Union{Vector,Compat.Nothing}
         end
-        ## constructor with optional keyword arguments, defaulting to Nullables
-        $contrastType(;
-                      base=Nullable{Any}(),
-                      levels=Nullable{Vector}()) =
-                          $contrastType(nullify(base),
-                                        nullify(levels))
+        ## constructor with optional keyword arguments, defaulting to nothing
+        $contrastType(; base=nothing, levels=nothing) = $contrastType(base, levels)
     end
 end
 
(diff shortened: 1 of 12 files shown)


ararslan:
The changes here look good to me so far. What is it that necessitates the dependency on Nullables?

kleinschmidt:
There are fields on the contrast types that are un-initialized by default.  Could probably replace that with `Union{T,Nothing}` though.

src/contrasts.jl
```diff
@@ -162,8 +162,8 @@ function ContrastsMatrix(contrasts::AbstractContrasts, levels::AbstractVector)
     # find index of base level. use contrasts.base, then default (1).
     baseind = isnull(contrasts.base) ?
               1 :
-              findfirst(equalto(get(contrasts.base)), c_levels)
-    if baseind < 1
+              Compat.findfirst(equalto(get(contrasts.base)), c_levels)
```
nalimilan: Should now use `isequal` instead of `equalto`.

src/modelmatrix.jl
```diff
@@ -62,7 +62,7 @@ function modelmat_cols(::Type{T},
     ## contrast matrix
     reindex = [findfirst(equalto(l), contrast.levels) for l in CategoricalArrays.index(v.pool)]
     contrastmatrix = convert(T, contrast.matrix)
-    return indexrows(contrastmatrix, reindex[v.refs])
+    return indexrows(contrastmatrix, Vector{Int}(reindex[v.refs]))
```
nalimilan: Better directly mark the comprehension as `Int` above?

kleinschmidt: you'd think so but it can contain `nothing`. (if not all levels of the categorical vector are contained in the contrast values). the test that hit this is the one where you make a model frame from a sub-dataframe.

kleinschmidt: in that case you don't even _hit_ the `nothing` values when you index with `v.refs`, and `ContrastMatrix` constructor uses unique values (rather than levels) so this is safe (or rather, an error would be detected earlier).

kleinschmidt: maybe use `coalesce` instead, on the comprehension? I feel like I saw a warning to that effect.

nalimilan: `coalesce` wouldn't throw an error in the presence of `nothing`, so a type assertion would be better.

We could also use `indexin`, that could be faster with many levels, but it would require converting the result (though hopefully at some point this kind of conversion will stop copying the underlying data).

kleinschmidt:
AFAIK the test failures on 0.6 have something to do with putting the struct definitions in the stats model wrapper types tests outside the `@testset` and furthermore _that_ has something to do with defining methods for those types _within_ the `@testset`.  There's an illegal instruction interrupt whenever it hits a method on those types (e.g., the `predict` here is the thing that causes it: https://github.com/JuliaStats/StatsModels.jl/pull/53/files#diff-3304fd5467cdfd2fc74b336bea5c60ddR50, and it goes away when I define that predict method outside the `@testset`.

src/modelmatrix.jl
```diff
@@ -93,7 +94,7 @@ function droprandomeffects(trms::Terms)
     if !any(retrms)  # return trms unchanged
         trms
     elseif all(retrms) && !trms.response   # return an empty Terms object
-        Terms(Any[],Any[],Array{Bool}((0,0)),Array{Bool}((0,0)), Int[], false, trms.intercept)
+        Terms(Any[],Any[],zeros(Bool,0,0),zeros(Bool,0,0), Int[], false, trms.intercept)
```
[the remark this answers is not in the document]
kleinschmidt: does that make a difference if they're 0-length? I went with this just because it's more concise.

src/contrasts.jl
```diff
@@ -137,7 +137,7 @@ function ContrastsMatrix(contrasts::AbstractContrasts, levels::AbstractVector)
     #    better to filter data frame first
     # 3. contrast levels missing from data: would have empty columns, generate a
     #    rank-deficient model matrix.
-    c_levels = get(contrasts.levels, levels)
+    c_levels = contrasts.levels === nothing ? levels : contrasts.levels
```
ararslan: You could write this as
```julia
c_levels = coalesce(contrasts.levels, levels)
```

src/contrasts.jl
```diff
@@ -199,24 +199,17 @@ function termnames(C::AbstractContrasts, levels::AbstractVector, baseind::Intege
...
 # Making a contrast type T only requires that there be a method for
 # contrasts_matrix(T,  baseind, n) and optionally termnames(T, levels, baseind)
 # The rest is boilerplate.
 for contrastType in [:DummyCoding, :EffectsCoding, :HelmertCoding]
     @eval begin
         mutable struct $contrastType <: AbstractContrasts
-            base::Nullable{Any}
-            levels::Nullable{Vector}
+            base::Any
+            levels::Union{Vector,Compat.Nothing}
```
ararslan: The `Compat.` prefix for `Nothing` is unnecessary, since it's exported.

src/modelmatrix.jl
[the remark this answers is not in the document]
ararslan: Yeah I'd think it shouldn't matter if they're 0-length.

src/contrasts.jl
kleinschmidt: cooool

src/contrasts.jl
```diff
@@ -160,11 +160,11 @@ function ContrastsMatrix(contrasts::AbstractContrasts, levels::AbstractVector)
     end
 
     # find index of base level. use contrasts.base, then default (1).
-    baseind = isnull(contrasts.base) ?
+    baseind = contrasts.base === nothing ?
```
Nosferican: How about `contrasts.base isa Nothing`?

nalimilan: In general it gives the same code. I think the exception is when the return type depends on this condition, in which case `isa` is currently the only way to make it inferrable.

PR #55, opened 2018-04-16
unique from categoricalarrays 0.3.8

kleinschmidt:
CategoricalArrays 0.3.8 changes `unique` to return values in the order they are encountered, rather than the natural (sorted) order.  This uses `intersect(levels(x), unique(x))` to impose the order of levels on the values returned by `unique`.


nalimilan:
https://github.com/JuliaLang/METADATA.jl/pull/14303

PR #57, opened 2018-04-27
WIP replace DataFrames with DataStreams Data.Table

kleinschmidt:
This introduces a marginally more backend-agnostic interface using DataStreams as a replacement for DataFrames: anything that can stream to a `Data.Table` (a named tuple of columns) can be used to make a ModelFrame.

Not all test pass yet (there's something funny with contrasts eltypes at least) but the basic functionality is there.

I copied some of the `completecases` code from DataFrames, but that's basically it. There's some ugliness because of how named tuples are constructed in 0.6 and 0.7, but it's pretty minimal.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
 julia 0.6
-DataFrames 0.11.5
+DataStreams
 StatsBase 0.20.1
 Compat 0.63
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
 julia 0.6
-DataStreams
+DataStreams 0.3.0
 StatsBase 0.20.1
 Compat 0.63
--- a/src/modelframe.jl
+++ b/src/modelframe.jl
@@ -47,7 +47,7 @@ julia> mf = ModelFrame(y ~ 1 + x, df)
 
 """
 mutable struct ModelFrame
-    df::AbstractDataFrame
+    dt::Data.Table
     terms::Terms
     msng::BitArray
     ## mapping from df keys to contrasts matrices
@@ -67,7 +67,7 @@ is_categorical(::AbstractArray) = true
 ##
 ## This modifies the Terms, setting `trms.is_non_redundant = true` for all non-
 ## redundant evaluation terms.
-function check_non_redundancy!(trms::Terms, df::AbstractDataFrame)
+function check_non_redundancy!(trms::Terms, d::Data.Table)
 
     (n_eterms, n_terms) = size(trms.factors)
 
@@ -81,7 +81,7 @@ function check_non_redundancy!(trms::Terms, df::AbstractDataFrame)
         for i_eterm in 1:n_eterms
             ## only need to check eterms that are included and can be promoted
             ## (e.g., categorical variables that expand to multiple mm columns)
-            if Bool(trms.factors[i_eterm, i_term]) && is_categorical(df[trms.eterms[i_eterm]])
+            if Bool(trms.factors[i_eterm, i_term]) && is_categorical(d[trms.eterms[i_eterm]])
                 dropped = trms.factors[:,i_term]
                 dropped[i_eterm] = 0
 
@@ -119,9 +119,9 @@ end
 ## Combine actual DF columns and contrast types if necessary to compute the
 ## actual contrasts matrices, levels, and term names (using DummyCoding
 ## as the default)
-function evalcontrasts(df::AbstractDataFrame, contrasts::Dict = Dict())
+function evalcontrasts(d::Data.Table, contrasts::Dict = Dict())
     evaledContrasts = Dict()
-    for (term, col) in eachcol(df)
+    for (term, col) in zip(fieldnames(d), d)
         is_categorical(col) || continue
         evaledContrasts[term] = ContrastsMatrix(haskey(contrasts, term) ?
                                                 contrasts[term] :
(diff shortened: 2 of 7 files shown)


src/modelframe.jl
```diff
@@ -131,31 +131,67 @@ function evalcontrasts(df::AbstractDataFrame, contrasts::Dict = Dict())
...
+function _nonmissing!(res, col::CategoricalArray{>: Missing})
+    for (i, el) in enumerate(col.refs)
+        res[i] &= el > 0
+    end
+end
+
+# handle some inconsistencies between how named tuples are constructed
+if isdefined(Core, :NamedTuple)
+    _select(d::Data.Table, cols::NTuple{N,Symbol} where N) = NamedTuple{cols}(d)
+    _filter(d::T, rows) where T<:Data.Table = T(([col[rows] for col in d]...))
```
nalimilan: `T((col[rows] for col in d)...)` should avoid an allocation.

src/modelframe.jl
```diff
@@ -131,31 +131,67 @@ function evalcontrasts(df::AbstractDataFrame, contrasts::Dict = Dict())
...
+    check_non_redundancy!(trms, d)
 
-    ModelFrame(df, trms, msng, evaledContrasts)
+    ModelFrame(d, trms, msng, evaledContrasts)
 end
 
-ModelFrame(df::AbstractDataFrame, term::Terms, msng::BitArray) = ModelFrame(df, term, msng, evalcontrasts(df))
-ModelFrame(f::Formula, d::AbstractDataFrame; kwargs...) = ModelFrame(Terms(f), d; kwargs...)
-ModelFrame(ex::Expr, d::AbstractDataFrame; kwargs...) = ModelFrame(Formula(ex), d; kwargs...)
+ModelFrame(f::Formula, d; kwargs...) = ModelFrame(Terms(f), d; kwargs...)
```
nalimilan: Any reason not to mark `d::Data.Table`? In general I prefer being explicit about what types are accepted, it makes errors clearer and helps documenting the function.

kleinschmidt: All of these constructors end up at one that does specify `Data.Table`.  My idea was to accept anything and let DataStreams sort out whether it could be streamed to a `Table`.  There's not a type restriction that specifies that, so we have to take Any.  The alternative would be to restrict the `ModelFrame` constructors to `Data.Table`, but allow anything in the `fit` methods.  I'd prefer this way to that, because then package authors who are constructing model frames manually also get the benefits of full DataStreams support without having to know about it.

REQUIRE
```diff
@@ -1,4 +1,4 @@
 julia 0.6
-DataFrames 0.11.5
+DataStreams
```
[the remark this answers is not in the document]
quinnj: agreed.

src/modelframe.jl
```diff
@@ -131,31 +131,67 @@ function evalcontrasts(df::AbstractDataFrame, contrasts::Dict = Dict())
...
+# handle some inconsistencies between how named tuples are constructed
+if isdefined(Core, :NamedTuple)
+    _select(d::Data.Table, cols::NTuple{N,Symbol} where N) = NamedTuple{cols}(d)
+    _filter(d::T, rows) where T<:Data.Table = T(([col[rows] for col in d]...))
+else
+    _select(d::Data.Table, cols::NTuple{N,Symbol} where N) = d[cols]
+    _filter(d::T, rows) where T<:Data.Table = T([col[rows] for col in d]...)
+end
+
+_size(d::Data.Table) = size(Data.schema(d))
```
quinnj: This could be a bit faster to just copy the definitions in DataStreams, but this is also fine.

kleinschmidt:
Okay the issue is that `Data.stream!(df, Data.Table)` converts `CategoricalArray`s to `Array{CategoricalValue}`s.  Is that behavior possible to change, @quinnj ?  Or is there some kind of work-around we could use here?

davidanthoff:
This would break the existing [IterableTables.jl](https://github.com/davidanthoff/IterableTables.jl) integration. Right now that provides compatibility with 20 different source types. Some of them would continue to work via the [DataStreams.jl](https://github.com/JuliaData/DataStreams.jl) integration instead, but I think there would still be about 15 source types that would stop working, including [Query.jl](https://github.com/davidanthoff/Query.jl).

The problem here is that [IterableTables.jl](https://github.com/davidanthoff/IterableTables.jl) is engaging in [type piracy](https://github.com/davidanthoff/IterableTables.jl/blob/master/src/integrations/statsmodels-missing.jl) to get this integration working, and this PR adds the same methods to ``ModelFrame`` (and maybe ``fit``? I haven't looked that carefully yet) that I've added in [IterableTables.jl](https://github.com/davidanthoff/IterableTables.jl).

I'm just slowly coming back to some julia work after a couple of weeks of university stuff, so haven't really thought about what could be done, just wanted to flag it for now.

nalimilan:
> Okay the issue is that Data.stream!(df, Data.Table) converts CategoricalArrays to Array{CategoricalValue}s. Is that behavior possible to change, @quinnj ? Or is there some kind of work-around we could use here?

@kleinschmidt We discussed that in the context of DataFrames some time ago. IIRC Jacob said that it would have to be handled by DataFrames.

@davidanthoff This PR is a good occasion to sort this system out so that type piracy isn't needed anymore. It looks like instead of defining fallbacks for `ModelFrame`, IterableTables could provide fallbacks for `Data.stream!(d, Data.Table)` for `d` types it supports. Do you think that would work?

kleinschmidt:
Superseded by #71

PR #60, opened 2018-05-18
delegate coefnames for wrapped models to model frame

kleinschmidt:
This is a StatsBase API method that was missing.

fixes JuliaStats/GLM.jl#235


kleinschmidt:
The alternative way to implement this is to delegate it to the wrapped model itself, and leave it up to the model packages to do something sensible (like use the rownames of the coeftable).  That might be slightly safer in cases where the model frame coefnames are not identical to the coeftable rownames, but I can't think of such a case off the top of my head...

PR #68, opened 2018-06-27
fixes copy(::Formula) when rhs is a single Symbol

AsafManela:
Added test on line 131 of test/formula.jl would fail, but now passes.


AsafManela:
done

PR #70, opened 2018-07-02
 msng -> nonmissing

matthieugomez:
msng had a confusing name and a wrong definition in doc. I have renamed the vector and corrected the doc.


kleinschmidt:
Thanks!

PR #71, opened 2018-08-07
Terms 2.0: son of Terms

kleinschmidt:
This is a pretty major re-thinking of how to represent terms in a formula.  It builds on #4, #54, and #57.  The basic idea is that the `@formula` macro lowers a formula expression to an expression where symbols are "wrapped" in a `Term` struct, and overloads operators like `+`, `&`, and `~` with methods of `Term`s that generate higher-order terms like interactions.  Additionally, this PR includes a mechanism by which calls to functions that don't have special meanings in the formula DSL are lowered to a call to `capture_call` which gets the original function called, the original expression, and an anonymous function that "wraps" that call.  The default result of that function is that it passes these onto a `FunctionTerm` constructor, but in principle package authors could intercept things at this point.  Whether or not a call is considered 'special' is also customizable, dispatching on `is_special(Val(::Symbol))`.

The other major new component is a schema representation that I "borrowed" from JuliaDB.ML.  Schemas are computed from a namedtuple of vectors (e.g., what DataStreams calls a Data.Table), and when applied to a formula will replace leaf `Term`s with `Categorical`/`ContinuousTerm`s.

The major _conceptual_ difference is that any subtype of `AbstractTerm` can generate model matrix columns, and the way that columns are combined to make higher order model matrices is handled by dispatch.  This provides, I think, much more flexibility in how package authors can "plug into" the formula pipeline, because they are no longer restricted to using fully-formed `ModelFrame`s.

That being said, I've tried to keep the ModelFrame/ModelMatrix structure for now to make it easier to see how things have changed.  I'd also like to consider how we actually use these structures to generate and fit models (e.g. #32).  But that's orthogonal enough that this is worth considering as is.

This is work in progress and I haven't even tried to get the tests passing yet because I wanted to talk about this at juliacon.  But I think it's close enough to the kind of structure I've had in mind for a long time that it's worth considering.

PR #76, opened 2018-09-29
Fix predict for drop_intercept models

AsafManela:
For models M that specify drop_intercept(M) = true, the fit(M,Formula,df) function adds back and then drops the intercept just before creating the model frame. This same idiom also needs to be applied to predict() otherwise contrast variables are mishandled, resulting in a dimension mismatch error.
This PR:
- fixes predict with drop_intercept model and categorical variables bug
- add tests that used to fail but now pass

### DIFF
--- a/src/statsmodel.jl
+++ b/src/statsmodel.jl
@@ -48,12 +48,12 @@ end
 """
     drop_intercept(::Type)
 
-Define whether a given model automatically drops the intercept. Return `false` by default. 
-To specify that a model type `T` drops the intercept, overload this function for the 
+Define whether a given model automatically drops the intercept. Return `false` by default.
+To specify that a model type `T` drops the intercept, overload this function for the
 corresponding type: `drop_intercept(::Type{T}) = true`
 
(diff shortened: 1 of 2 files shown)


nalimilan:
Cc: @piever

piever:
LGTM! At the time I forgot to test `predict` for the model with no intercept, well spotted!

AsafManela:
This has been reviewed a long time ago. Why not merge it?

kleinschmidt:
On the one hand, everything is going to change once #71 is merged (which should be soon, inshallah) in a way that will make this redundant.  On the other hand, it might make sense to merge this one now in order to fix an obvious bug instead of having to backport it... thoughts @nalimilan ?

nalimilan:
I'd say merge, unless it creates conflicts that are too costly for you to fix (unlikely).

kleinschmidt:
Agreed.

kleinschmidt:
Thanks for the fix, and sorry for the delay in merging!

AsafManela:
No worries. Thanks for merging. I look forward to #71

PR #79, opened 2019-01-16
Explicitly ask for column names in evalcontrasts when calling eachcol.

andreasnoack:
The removes a warning which broke the GLM doctests. I plan to make a bugfix release once this is in.


andreasnoack:
It looks like the doctests are failing but given the expected arrival of #71, I don't think it's worth fixing them.

PR #77, opened 2018-11-11
predict_confidence

mkborregaard:
Handle prediction that returns a confidence interval
cc @nalimilan
cf https://github.com/JuliaStats/GLM.jl/pull/253

PR #80, opened 2019-01-17
Revert "Handle prediction that returns a confidence interval (#77)"

andreasnoack:
This reverts commit a64cb50f2e8cfc0e8718fd7651458cd15373d934.

As discussed in #77. Once this has been merged, I'll make a new release and should be able to finish the `predict` behavior without any rush. (@nalimilan we already require DataFrames 0.15 which includes the two argument `eachcol` methods, see https://github.com/JuliaData/DataFrames.jl/commit/dcac4bc383231388629f25877e7b1304cffe0632.)

### DIFF
--- a/src/statsmodel.jl
+++ b/src/statsmodel.jl
@@ -48,12 +48,12 @@ end
 """
     drop_intercept(::Type)
 
-Define whether a given model automatically drops the intercept. Return `false` by default.
-To specify that a model type `T` drops the intercept, overload this function for the
+Define whether a given model automatically drops the intercept. Return `false` by default. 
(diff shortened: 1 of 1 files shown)


nalimilan:
> (@nalimilan we already require DataFrames 0.15 which includes the two argument `eachcol` methods, see [JuliaData/DataFrames.jl@dcac4bc](https://github.com/JuliaData/DataFrames.jl/commit/dcac4bc383231388629f25877e7b1304cffe0632).)

Yes we require it because I committed that change directly on your PR before merging. :-)

PR #81, opened 2019-01-17
accept namedtuples and return a DataFrame

mkborregaard:
As discussed on Slack. Some considerations:

1. I have made it accept both a Matrix or a NamedTuple. That makes it independent of the GLM version. I don't know if that's desirable but it was suggested on the slack.
2. I have made it return a DataFrame for now as that is what it accepts. We can edit for Tables when Terms 2.0 is merged I guess?
3. I have used dispatch on a nonexported helper function rather than testing the types with if statements. Not sure what the style is here.
4. There is no single-argument `predict(mymodel, interval = :confidence)` method. Where should that be added, here or in GLM?

### DIFF
--- a/src/statsmodel.jl
+++ b/src/statsmodel.jl
@@ -48,12 +48,12 @@ end
 """
     drop_intercept(::Type)
 
-Define whether a given model automatically drops the intercept. Return `false` by default. 
-To specify that a model type `T` drops the intercept, overload this function for the 
+Define whether a given model automatically drops the intercept. Return `false` by default.
+To specify that a model type `T` drops the intercept, overload this function for the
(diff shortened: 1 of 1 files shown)


nalimilan:
> * There is no single-argument `predict(mymodel, interval = :confidence)` method. Where should that be added, here or in GLM?

It should live here (since GLM doesn't know which rows contained missing values and were skipped), but probably better wait until https://github.com/JuliaStats/StatsModels.jl/pull/71 is merged.

nalimilan:
OK to merge?

mkborregaard:
Yes from me!

nalimilan:
I've tagged 0.5.0: https://github.com/JuliaLang/METADATA.jl/pull/20913

PR #89, opened 2019-03-11
mention that full-dummy is the same as one-hot in ML literature.

oxinabox:
minor docs change.
I am reading the docs and trying to relate things to things that I know


kleinschmidt:
Thanks!  I'd actually meant to mention that but it got lost in the shuffle.

PR #91, opened 2019-03-18
Bug-fix for respecting levels

Nosferican:
Allows contrasts to be aware of the order (especially important for ordinal features).

### DIFF
--- a/test/contrasts.jl
+++ b/test/contrasts.jl
@@ -82,16 +82,11 @@
                                 1  0  1]
     @test coefnames(mf) == ["(Intercept)"; "x: c"; "x: b"]
     
-    # Ordinal factors
+    # respect order of levels
     
-    data = DataFrame(x = levels!(categorical(['A', 'B', 'C', 'C'],
-                                             ordered = true),
-                                 ['C', 'B', 'A']))
-    f = @formula(x ~ 1)
-    f = apply_schema(f, schema(data))
-    y = modelcols(f.lhs, data)
-    f = apply_schema(f, schema(data, Dict(:x => DummyCoding(levels = levels(data.x)))))
-    @test == y = modelcols(f.lhs, data)
+    data = DataFrame(x = levels!(categorical(['A', 'B', 'C', 'C', 'D']), ['C', 'B', 'A', 'D']))
+    f = apply_schema(@formula(x ~ 1), schema(data))
+    @test modelcols(f.lhs, data) == [0 1 0; 1 0 0; 0 0 0; 0 0 0; 0 0 1]
 
     # Helmert coded contrasts
     setcontrasts!(mf, x = HelmertCoding())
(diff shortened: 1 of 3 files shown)


test/contrasts.jl
```diff
@@ -81,6 +81,17 @@
...
+    
+    # Ordinal factors
+    
+    data = DataFrame(x = levels!(categorical(['A', 'B', 'C', 'C'],
+                                             ordered = true),
+                                 ['C', 'B', 'A']))
+    f = @formula(x ~ 1)
+    f = apply_schema(f, schema(data))
+    y = modelcols(f.lhs, data)
+    f = apply_schema(f, schema(data, Dict(:x => DummyCoding(levels = levels(data.x)))))
```
nalimilan: Passing `levels` here is redundant and defeats the point of using a categorical array, right?

test/contrasts.jl
```diff
@@ -81,6 +81,17 @@
                                 1 -1 -1
                                 1  0  1]
     @test coefnames(mf) == ["(Intercept)"; "x: c"; "x: b"]
+    
+    # Ordinal factors
+    
+    data = DataFrame(x = levels!(categorical(['A', 'B', 'C', 'C'],
+                                             ordered = true),
```
nalimilan: Better not set `ordered=true`, as the behavior should be the same for unordered arrays. Then also change "ordinal" to "categorical" above.

test/contrasts.jl
Nosferican: It is redundant now, but works as a test since it errors with current master. That is just to verify that it is working as if/intended.

nalimilan: OK but then below you need to check that `y == modelcols(f.lhs, data) == somereference` (or call `@test y == somereference` twice).

nalimilan: Ah, and also better pass levels in a different order from `levels` and from the order of appearance, to ensure it really has an effect.

nalimilan: And add an unused value to `levels` to check that it's dropped.

Nosferican: That is what it is doing, it is first computing the value and storing it as `y` and the I test the value against the reference which is `modelcols(f.lhs, data)` with the formula now been updated with an explicit order by contrasts.

Nosferican: The test for unused levels was already included in the `modelmatrix.jl`. I put a comment on that test that it also checks correctness against non-occurring levels.

nalimilan: > That is what it is doing, it is first computing the value and storing it as `y` and the I test the value against the reference which is `modelcols(f.lhs, data)` with the formula now been updated with an explicit order by contrasts.

Yes but this doesn't test the fact that passing levels to `DummyCoding` has any effect. Better check against an explicit expected array.

Nosferican:
Let me know if there is anything else or if approved.

kleinschmidt:
Thanks!

PR #98, opened 2019-04-05
Remove stray comment in the docs

oxinabox:


kleinschmidt:
IIRC this comment is actually necessary so that the code block and be interpreted as a doctest...

kleinschmidt:
(which in turn is required so that the types/methods defined in it can be used in later doctest blocks)

kleinschmidt:
https://travis-ci.org/JuliaStats/StatsModels.jl/jobs/516250063#L692

oxinabox:
Really?
There must be a better way
It shows up in the content

kleinschmidt:
The only alternative as far as I could tell would be to duplicate the code in that block in a `@meta` block which defeats the purpose of a doctest. Or putting `julia> ` in front of every line in that block which...

PR #104, opened 2019-05-20
RIP REQUIRE, hello Project.toml

kleinschmidt:
In preparation for tagging a release, this replaces REQUIRE and test/REQUIRE with Project.toml

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,0 +1,22 @@
+name = "StatsModels"
+uuid = "3eaba693-59b7-5ba5-a881-562e759f1c8d"
+
+[deps]
+CategoricalArrays = "324d7699-5711-5eae-9e2f-1d82baa6b597"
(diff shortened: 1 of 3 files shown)


ararslan:
Needs a version underneath uuid in order to be tagged, otherwise LGTM if CI agrees

kleinschmidt:
And just to be clear, merging a Project.toml with a version is **not** enough to trigger the registration bot right?  THat requires a manual step?

ararslan:
Correct. You do at-JuliaRegistrator `register()` and it uses the version in Project.toml to make a PR against the General registry, then once that's merged, TagBot makes a tag for the version on the repo.

ararslan:
Still needs a version...

kleinschmidt:
well I'm a moron. time to make another PR like an idiot

PR #106, opened 2019-05-21
just copy continuous terms in modelcols, don't convert to float64

kleinschmidt:
Miraculously, tests pass with just this change.  The reason is that the underlying columns are all `hcat`ed together at the last stage, and for most of the normal cases there's an intercept term (or a non-intercept term), both of which are arrays with eltype Float64.

It might be possible to get away without even copying, just passing the underlying column straight through, as long as we do defensive copy at the very last stage (matrix term seems like a reasonable place to do this).  But that might be overkill.

This doesn't address the possibility that someone might want a model matrix that's not just Float64s.

As a note to self, I think the long-term solution is to create the model matrix in-place, which would require adding `modelcols!` or some such.  That would reduce copies to a minimum but be a bit more complex (the "setup" stage where you create the empty array of the desired type needs to be handled).

(partially) fixes JuliaStats/Survival.jl#10

### DIFF
--- a/src/terms.jl
+++ b/src/terms.jl
@@ -472,7 +472,7 @@ modelcols(ts::TupleTerm, d::NamedTuple) = modelcols.(ts, Ref(d))
 modelcols(ft::FunctionTerm{Fo,Fa,Names}, d::NamedTuple) where {Fo,Fa,Names} =
     ft.fanon.(getfield.(Ref(d), Names)...)
 
-modelcols(t::ContinuousTerm, d::NamedTuple) = Float64.(d[t.sym])
+modelcols(t::ContinuousTerm, d::NamedTuple) = copy.(d[t.sym])
 
 modelcols(t::CategoricalTerm, d::NamedTuple) = t.contrasts[d[t.sym], :]
 


nalimilan:
Makes sense. What's the point of copying? AFAICT a copy will be made anyway to create the matrix, right?

kleinschmidt:
Initially I was just trying to find something that would have the same semantics as the original `Float64.(col)`.  I think the once place where copying is important is when there's no other terms combined so nothing is copied (e.g., the LHS of a formula).  In those cases, a fitting algorithm might overwrite the underlying data if it thinks it can modify the `y` input.

nalimilan:
What's the best place to make a copy then? Is there a way to ensure we don't make a copy for terms that end up in a matrix?

kleinschmidt:
No there's no way to ensure that at the moment.  I think the best way to do that is to add in-place `modelcols!` methods and call that from `modelcols(::MatrixTerm)`.  That's a feature I'm planning to add but haven't gotten to yet.

Then we could have the non-modifying `modelcols` copy for continuous terms since the only time it'll be called is when there's no matrix term around it.

kleinschmidt:
Another possibility would be to also wrap the left-hand side in a `MatrixTerm` too.  This would have the downside of requiring packages to be able to consume an Nx1 matrix for `y` instead of a vector but would really simplify this (wouldn't require copy).

kleinschmidt:
It would also address #110

PR #108, opened 2019-06-11
Add Basic Lag and Lead terms

oxinabox:


src/temporal_terms.jl
```diff
@@ -0,0 +1,56 @@
...
+    term = apply_schema(term_parsed, sch, mod)
+
+    (nsteps_parsed isa ConstantTerm) ||
+        throw(ArgumentError("Lag step must be a number (got $step_parsed)"))
+
+    nsteps = nsteps_parsed.n
+    return LagTerm(term, nsteps)
+end
+
+function modelcols(lag::LagTerm, d::NamedTuple)
```
kleinschmidt: this will only work when `d` is a "column table" (e.g. named tuple of vectors), not for single-row namedtuples.  so better to specialize on `Tables.ColumnTable`, and throw an error that says that lag is only supported for in-memory data.

test/temporal_terms.jl
```diff
@@ -0,0 +1,23 @@
+using StatsModels
+using StatsBase
+using StatsModels: width
+using DataStructures
+
+@testset "Temporal Terms" begin
```
kleinschmidt: add tests for categorical term?

src/temporal_terms.jl
oxinabox: Docs for extending need updating

oxinabox: It would be nice to support those for row tables.
I guess via a buffer in the term,
That gets iterated each time a row is processed.
But needs to be reset whenever one starts again.

Where can I find a example of an opt that works on row tables?

src/temporal_terms.jl
```diff
@@ -0,0 +1,56 @@
+# Basic Lag:
+
+# syntax:
+"""
+    lag(term, nsteps)
+This `@formula` term is used to introduce lagged variables.
+For example `lag(x,1)` effectively adds a new column containing
+the value of the `x` column from the previous row.
+If there are no such row (e.g. because this is the first row), then the lagged column will
```
nalimilan:
```suggestion
If there is no such row (e.g. because this is the first row), then the lagged column will
```

src/temporal_terms.jl
```diff
@@ -0,0 +1,56 @@
+# Basic Lag:
+
+# syntax:
+"""
+    lag(term, nsteps)
```
nalimilan:
```suggestion
    lag(term, nsteps::Integer)

```

BTW, it seems standard to have `nsteps=1` by default (e.g. dplyr, ShiftedArrays...).

src/temporal_terms.jl
```diff
@@ -0,0 +1,56 @@
...
+This `@formula` term is used to introduce lagged variables.
+For example `lag(x,1)` effectively adds a new column containing
+the value of the `x` column from the previous row.
+If there are no such row (e.g. because this is the first row), then the lagged column will
+contain `missing`.
+
+Note: this is only a basic row-wise lag operation.
+It is up to the user to ensure that that there data is sorted by the temporal variable,
```
nalimilan:
```suggestion
It is up to the user to ensure that data is sorted by the temporal variable,
```

src/temporal_terms.jl
```diff
@@ -0,0 +1,56 @@
...
+and that their observations are spaced with regular time-steps.
```
nalimilan:
```suggestion
and that observations are spaced with regular time-steps.
```
("the user" is singular)

src/temporal_terms.jl
```diff
@@ -0,0 +1,56 @@
...
+    n_cols = size(original_cols, 2)
+    padding = fill(missing, (lag.nsteps, n_cols))
+    padded_cols = [padding; original_cols]
+    return padded_cols[1:size(original_cols, 1), :]
+end
+
+width(lag::LagTerm) = width(lag.term)
+
+function StatsBase.coefnames(lag::LagTerm)
+    # Possible Hack: Reset memory here.
```
nalimilan: What does this mean?

src/temporal_terms.jl
```diff
@@ -0,0 +1,56 @@
...
+    return coefnames(lag.term) .* "_lagged_by_$(lag.nsteps)"
```
nalimilan: This sounds really long. How about something like `_lag$(lag.nsteps)`?

src/temporal_terms.jl
```diff
@@ -0,0 +1,56 @@
...
+
+Note: this is only a basic row-wise lag operation.
+It is up to the user to ensure that that there data is sorted by the temporal variable,
+and that their observations are spaced with regular time-steps.
+(Which may require adding extra-rows filled with `missing` values.)
+"""
+lag(term, nsteps) = error("`lag`  should only be used within a @formula");
+
+# type of model where syntax applies:
+const LAG_CONTEXT = Any
```
nalimilan: I don't see the point of this. Why not use `Any` directly?

src/StatsModels.jl
```diff
@@ -39,8 +39,11 @@ export
     FormulaTerm,
     InterceptTerm,
     FunctionTerm,
+    LagTerm,
     MatrixTerm,
 
+    lag,
```
nalimilan: ShiftedArrays [also exports `lag`](https://github.com/piever/ShiftedArrays.jl#shifting-the-data). Not sure what we can do about it, but ideally `lag(x, n)` would work both in a formula and in the normal environment. Maybe we should define the function in DataAPI.jl? Cc @piever

src/temporal_terms.jl
oxinabox: ok

src/temporal_terms.jl
oxinabox: Oh that should be deleted, leftover note from when I was not understanding how `modelcols` worked

```suggestion
```

src/temporal_terms.jl
oxinabox: This is basically copied from the docs on extending.
I think it has some merit, but i can remove it if you think that makes sense to

src/StatsModels.jl
oxinabox: We could just add ShiftedArrays as a dependency,
then use it in the implementation -- I think it would be cleaner than the current method?
and reexport the name `lag`

We don't have a conflicting definition for it

src/temporal_terms.jl
nalimilan: Well, that's useful in the docs, but since it's `Any` in this case it doesn't add anything IMO.

src/StatsModels.jl
nalimilan: Adding a dependency to an essential package is never great, but maybe ShiftedArrays is small enough that it's OK.

src/temporal_terms.jl
```diff
@@ -0,0 +1,56 @@
...
+Base.show(io::IO, lag::LagTerm) = print(io, "lag($(lag.term), $(lag.nsteps))")
+
+function apply_schema(t::FunctionTerm{typeof(lag)}, sch, mod::Type{<:LAG_CONTEXT})
+    @assert length(t.args_parsed) == 2
+    term_parsed, nsteps_parsed = t.args_parsed
+
+    term = apply_schema(term_parsed, sch, mod)
+
+    (nsteps_parsed isa ConstantTerm) ||
+        throw(ArgumentError("Lag step must be a number (got $step_parsed)"))
```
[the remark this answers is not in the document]
oxinabox:
```suggestion
        throw(ArgumentError("Lag step must be a number (got $nstep_parsed)"))
```

src/temporal_terms.jl
```diff
@@ -0,0 +1,56 @@
...
+# type of model where syntax applies:
+const LAG_CONTEXT = Any
+
+# struct for behavior
+struct LagTerm{T<:AbstractTerm} <: AbstractTerm
+    term::T
+    nsteps::Int
+end
+
+Base.show(io::IO, lag::LagTerm) = print(io, "lag($(lag.term), $(lag.nsteps))")
```
oxinabox: is this the right way to do this?
Does it it render right for things like categorical variables?
Should I be doing something with mimetypes?

nalimilan: That should be OK AFAICT.

oxinabox:
Should add a test for
```
@formula y ~ lag(lag(x, 1), 2)
```

src/temporal_terms.jl
oxinabox: how about `⌛-($lag.nsteps)`

src/StatsModels.jl
oxinabox: I'm going to do this, and we can see what it looks like

src/temporal_terms.jl
kleinschmidt: Yeah in the docs it's meant to be pedagogical.

src/temporal_terms.jl
kleinschmidt: There isn't one currently (at least not one like this, where the behavior is going to be different). @nalimilan and I talked a bit about using the idea of "stateful transforms" from patsy (the formula library behind python statsmodels): https://patsy.readthedocs.io/en/latest/stateful-transforms.html

But haven't looked into that in any depth.

oxinabox: I feel like it might be not too crazy
to have
`modelrows(lag::LagTerm, lag_term_state, row::NamedTuple, row_iterator_start)`

Then this could be implemeted witha `CircularBuffer`

nalimilan: Maybe better start with support for column tables only, and discuss row tables separately. That way we can focus on the design here.

oxinabox: Indeed

src/StatsModels.jl
piever: I think it should be the same function. ShiftedArrays is quite small and I suspect the implementation can be useful, not just the name, so maybe it's easier to just take the dependency? I was planning to remove the two dependencies ShiftedArrays currently has in the near future.

oxinabox:
All is done.

I also added a test for implementing diff.
Which is broken because of #114

src/temporal_terms.jl
```diff
@@ -0,0 +1,60 @@
...
+    then the $opped column will contain `missing` for that entry.
+
+    Note: this is only a basic row-wise $op operation.
+    It is up to the user to ensure that data is sorted by the temporal variable,
+    and that observations are spaced with regular time-steps.
+    (Which may require adding extra-rows filled with `missing` values.)
+"""
+
+@doc lead_lag_docstring("lag", "lagged", "previous", "first") lag
+@doc (lead_lag_docstring("lead", "lead", "next", "last")) lead
```
nalimilan:
```suggestion
@doc lead_lag_docstring("lead", "lead", "next", "last") lead
```

src/temporal_terms.jl
```diff
@@ -0,0 +1,60 @@
...
+
+
+# struct for behavior
+struct LeadLagTerm{T<:AbstractTerm, F<:Union{typeof(lead), typeof(lag)}} <: AbstractTerm
+    term::T
+    nsteps::Int
+end
+
+op(::LeadLagTerm{<:Any, typeof(lag)}) = lag
```
nalimilan: Why not use `F.instance` as in `apply_schema`?

test/temporal_terms.jl
```diff
@@ -0,0 +1,105 @@
...
+    @testset "Lag" begin
+        @testset "Basic use" begin
+            df = (y=1:10, x = 1:10)
+            f = @formula(y ~ lag(x, 0) + lag(x, 1) + lag(x, 3) + lag(x, 11))
```
nalimilan: Should negative values work? Better test that too.

src/temporal_terms.jl
oxinabox: In an earlier version of this code `op` was called in a bunch of places,.
but not it is only called in 1 so that makes sense.

test/temporal_terms.jl
oxinabox: That is tested below

oxinabox:
This is ready for rereview

src/temporal_terms.jl
```diff
@@ -0,0 +1,58 @@
...
+        nsteps = nsteps_parsed.n
+    else
+        throw(ArgumentError("`$opname` terms require 1 or 2 arguments."))
+    end
+
+    term = apply_schema(term_parsed, sch, ctx)
+    return LeadLagTerm{typeof(term), F}(term, nsteps)
+end
+
+function modelcols(ll::LeadLagTerm{<:Any, F}, d::Tables.ColumnTable) where F
```
kleinschmidt: maybe add a method for `d::NamedTuple` that just throws an error so that it's clear that row-wise operation isn't supported?  (and tests for this...)

oxinabox: The MethodError will be clear enough I think?
Manually reimplementing MethodErrors is a bit of an antipattern.

nalimilan: Well the `MethodError` is going to be quite obscure for 90% of people who don't know what `LeadLagTerm` or `ColumnTable` (both unexported types) are.

src/temporal_terms.jl
```diff
@@ -0,0 +1,58 @@
...
+@doc lead_lag_docstring("lead", "lead", "next", "last") lead
+
+
+# struct for behavior
+struct LeadLagTerm{T<:AbstractTerm, F<:Union{typeof(lead), typeof(lag)}} <: AbstractTerm
+    term::T
+    nsteps::Int
+end
+
+opname(::LeadLagTerm{<:Any, typeof(lag)}) = "lag"
```
nalimilan: This can be removed too: it's used in a single place, where `string(F.instance)` would do.

oxinabox: `string` of that function gives a fully qualified name: `ShiftedArrays.lag`

src/temporal_terms.jl
oxinabox: TBH it will be worse as ColumnTable is an alias for a ugly parametric named tuple

Ok. I will introduce it.

src/temporal_terms.jl
nalimilan: Right. You can use `string(nameof(F.instance))` instead.

oxinabox: Right, but at that point, I lean towards just having a function. Like I currently have.

nalimilan: Then why not use it to compute `opname` in `apply_schema` too? (BTW, one error message uses backticks, and the other doesn't).

oxinabox:
Done.
Turns out row tables do infact work.
because I guess we convert everything into ColumnTables at the start.
Which is not great, but which is a seperate issue.
Test has been added to ensure they continue to work.

Docs have been added

docs/src/temporal_terms.md
```diff
@@ -0,0 +1,65 @@
...
+julia> f = apply_schema(f, schema(f, df))
+FormulaTerm
+Response:
+  y(continuous)
+Predictors:
+  x(continuous)
+  lag(x, 2)
+  lead(x, 2)
+
+julia> modelcols(f, df)
```
nalimilan: That line doesn't sound too useful. BTW, isn't there a simpler way to print a data frame than the next line?

oxinabox: `modelcols` is the primary way of accessing this package, so it seems fitting to include it.

As to printing a dataframe,
I am not aware of one.
I assume that `ModelFrame` does something like that,
but since it is schedualled to be remove probably do not want to use it

oxinabox: Can I halve more guidance in resolving this, so we can merge this todayish?

nalimilan: @kleinschmidt?

kleinschmidt: I'd suggest just using `modelmatrix` (which calls `modelcols` on the rhs).

kleinschmidt: Elsewhere in the docs we don't print the colnames so it's fine to just show the matrix here.

oxinabox: done

oxinabox:
bump

PR #118, opened 2019-06-17
Document to consider four times for internal transformation

oxinabox:
- Syntax
 - Schema
 - Semantics
 - Data

### DIFF
--- a/docs/src/internals.md
+++ b/docs/src/internals.md
@@ -119,7 +119,7 @@ Internally a lot of the work at syntax time is done by the `parse!` function.
 
 The next phase of life for a formula begins when a _schema_ for the data becomes
 available.  A schema is a mapping from data columns to a concrete term
-type---either a `ContinuousTerm` or a `CategoricalTerm`---which represents all
+type—either a `ContinuousTerm` or a `CategoricalTerm`—which represents all
 the summary information about a data column necessary to create a model matrix
 from that column.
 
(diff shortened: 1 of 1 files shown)


docs/src/internals.md
```diff
@@ -20,18 +20,20 @@ A formula goes through a number of stages, starting as an
...
 
-1. "Syntax time" when only the surface syntax is available
+1. "Syntax time" when only the surface syntax is available, during the defintion of `@formula`
 2. "Schema time" incorporates information about **data invariants** (types of each
    variable, levels of categorical variables, summary statistics for continuous
-   variables) and the **model type**.
-3. "Data time" when the actual data values themselves are available.
+   variables) and the over all structure of the **data**, during the invocation of `schema`
+3. "Semantics time" incorperates information about the **model type (context)**, and custom terms,
+during the call of `apply_schema`
```
nalimilan:
```suggestion
   during the call of `apply_schema`
```

docs/src/internals.md
```diff
@@ -20,18 +20,20 @@ A formula goes through a number of stages, starting as an
 expression that's passed to the `@formula` macro and ending up generating a
 numeric matrix when ultimately combined with a tabular data source:
 
-1. "Syntax time" when only the surface syntax is available
+1. "Syntax time" when only the surface syntax is available, during the defintion of `@formula`
```
nalimilan:
```suggestion
1. "Syntax time" when only the surface syntax is available, during the definition of `@formula`
```
But is "definition" the right word?

docs/src/internals.md
```diff
@@ -20,18 +20,20 @@ A formula goes through a number of stages, starting as an
...
 2. "Schema time" incorporates information about **data invariants** (types of each
    variable, levels of categorical variables, summary statistics for continuous
-   variables) and the **model type**.
-3. "Data time" when the actual data values themselves are available.
+   variables) and the over all structure of the **data**, during the invocation of `schema`
+3. "Semantics time" incorperates information about the **model type (context)**, and custom terms,
```
nalimilan:
```suggestion
3. "Semantics time" incorporates information about the **model type (context)**, and custom terms,
```

docs/src/internals.md
oxinabox: yes I think definition is the rigth word
I am defining an `foo = @formula(y ~ x) `

nalimilan: Precisely, I would say this is *calling* `@formula`. Defining it would be `macro formula...`.

oxinabox: Right, we are defining _the formula `foo`_ by calling the `@formula` macro.
Or to put it another way:
_The definition of the formula `foo` is written in the DSL that is processed by the `@formula` macro_

I am not sure on good compact wording.

nalimilan: I'd say "during the construction of the formula".

kleinschmidt:
I've pushed my doc updates as well. @oxinabox, if this looks good to you I'll
merge.

docs/src/internals.md
oxinabox:
```suggestion
1. "Syntax time" when only the surface syntax is available, when the `@formula` macro is invoked.
```

oxinabox:
LGTM
.
It is a clear improvement.
I think better still is possible,  but lets get this in for now

docs/src/internals.md
kleinschmidt: @nalimilan is "invoking `@formula`" okay with you?

kleinschmidt:
@nalimilan any final thoughts before we merge this?

docs/src/internals.md
```diff
@@ -163,13 +172,87 @@ Dict{Any,Any} with 2 entries:
...
+julia> sch1 = Dict(term(:a) => cont_a, term(:b) => cat_b)
+Dict{Term,AbstractTerm} with 2 entries:
+  a => a
+  b => b
+```
+
+#### Semi-automatic: data subsets
+
+A slightly more convenient method for generating a schema is provided by
+[`concrete_term`](@ref) the internal function which extracts invariants from a
```
nalimilan:
```suggestion
the [`concrete_term`](@ref) internal function, which extracts invariants from a
```

docs/src/internals.md
```diff
@@ -163,13 +172,87 @@ Dict{Any,Any} with 2 entries:
...
+data column and returns an concrete type.  This can be used to generate concrete
```
nalimilan:
```suggestion
data column and returns a concrete type.  This can be used to generate concrete
```

docs/src/internals.md
```diff
@@ -163,13 +172,87 @@ Dict{Any,Any} with 2 entries:
...
+available, which includes the schema of the data to be transformed as well as
+the _context_, or the type of model that will be fit.  This stage is implemented
+by [`apply_schema`](@ref).  Among other things, this _instantiates_ placeholder
 terms:
+
 * `Term`s become `ContinuousTerm`s or `CategoricalTerm`s
 * `ConstantTerm`s become `InterceptTerm`s
 * Tuples of terms become [`MatrixTerm`](@ref)s where appropriate to explicitly indicate
   they should be concatenated into a single model matrix
+* Any model (context) specific interperation of the terms is made, including
```
nalimilan:
```suggestion
* Any model-specific (context-specific) interpretation of the terms is made, including
```

PR #123, opened 2019-06-25
increase compatiliby to julia 1

kleinschmidt:


oxinabox:
Dup: #122 close one merge the other?

PR #127, opened 2019-07-03
Clean up tuples with sum after apply_schema

kleinschmidt:
This is necessary to support special terms that return a tuple of other terms.
This would also allow supporting `*` or even R's `^` via normal "special term"
mechanisms.


oxinabox:
Could do with a test that breaks without this and now passes

kleinschmidt:
Have another look @oxinabox?  I've added tests.

PR #137, opened 2019-08-01
Make modelcols(::Term, d) an error

### LINKED ISSUE #136: Stack overflow for `modelcols` before `apply_schema`
Reported by @JoshDay on slack:

``` julia
julia> using IndexedTables, StatsModels

julia> t = table((x=1:100,y=randn(100)));
 
julia> f = @formula(y ~ x)
FormulaTerm
Response:
  y(unknown)
Predictors:
  x(unknown)

julia> modelcols(f, t)
Bus error: 10
```

This is because there's no `modelcols` method for `::Term` so it just loops
forever in the method that handles converting general data types to
`ColumnTable`.

kleinschmidt:
This fixes #136 (stackoverflow when there's no more specific method than
`modelcols(::Any, ::Any)`).  An alternative would be to just return the
underlying column (which is probably what we'd want if we do something like
#117).  Making it an explicit error for now means that people can't rely on that
behavior until we've made a decision about what we want long-term.

One larger issue that I'm not sure how to address is that there's currently no
sensible fallback for `::AbstractTerm`s that don't have modelcols methods.
There shouldn't be any of those at this point but if there _are_ then they will
produce the same stack overflow as #136.  The problem is that when you add
methods like `modelcols(::AbstractTerm, ::Any)` or `modelcols(::AbstractTerm,
::NamedTuple)` you get ambiguity errors or problems when custom terms don't
specialize their `modelcols` methods on the data type.  A potential workaround
would be to put a check in the `modelcols(::Any, ::Any)` that checks if the
second argument is already a named tuple and throws an error if it is, but that
feels like kind of a hack.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StatsModels"
 uuid = "3eaba693-59b7-5ba5-a881-562e759f1c8d"
-version = "0.6.2"
+version = "0.6.3"
 
 [deps]
 CategoricalArrays = "324d7699-5711-5eae-9e2f-1d82baa6b597"
--- a/src/terms.jl
+++ b/src/terms.jl
@@ -468,6 +468,10 @@ julia> modelcols(MatrixTerm(ts), d)
 """
 modelcols(ts::TupleTerm, d::NamedTuple) = modelcols.(ts, Ref(d))
 
+modelcols(t::Term, d::NamedTuple) =
+    throw(ArgumentError("can't generate modelcols for un-typed term $t. " *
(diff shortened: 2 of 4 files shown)


ararslan:
> Making it an explicit error for now means that people can't rely on that
behavior until we've made a decision about what we want long-term.

:+1: to this in particular.

kleinschmidt:
Okay I've added a fallback guard (and a test using a dummy term type) in the `modelcols(::Any, ::Any)` method for now.

PR #139, opened 2019-08-02
One-sided formula, 2.0

kleinschmidt:
This supersedes #20 and provides support for one-sided formulae.  The required
changes were surprisingly minimal and non-disruptive (and eliminate nearly half
of the "broken" tests!).  The strategy is to put `nothing` on the LHS, which is
handled by a one-arg method.

``` julia
~(t::TermOrTerms) = ~(nothing, t)
```

And return `nothing` for `modelcols(::Nothing, d)`.  The only wart is you have
to do `@formula(~(1 + x + y))` to make sure that `~` parses correctly (the
precedence is different for prefix vs. infix).  The macro tweak from #20 could
fix that but I'm not sure it's worth it.

### DIFF
--- a/src/formula.jl
+++ b/src/formula.jl
@@ -59,7 +59,7 @@ The rules that are applied are
 """
 macro formula(ex)
     is_call(ex, :~) || throw(ArgumentError("expected formula separator ~, got $(ex.head)"))
-    length(ex.args) == 3 ||  throw(ArgumentError("malformed expression in formula $ex"))
+    length(ex.args) in (2,3) ||  throw(ArgumentError("malformed expression in formula $ex"))
     terms!(sort_terms!(parse!(ex)))
 end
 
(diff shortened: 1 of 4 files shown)


kleinschmidt:
On slack we discussed the possibility that `0 ~ stuff` might work just as well and require no changes at all (since it `0` is converted to `InterceptTerm{false}` during `apply_schema`, which generates an nx0 array with `modelcols`):

```julia
julia> apply_schema(term(0), StatsModels.Schema()) |> dump
InterceptTerm{false} 0

julia> modelcols(apply_schema(term(0), StatsModels.Schema()), (a=1,))
1×0 Array{Float64,2}

julia> modelcols(apply_schema(term(0), StatsModels.Schema()), (a=rand(10),))
10×0 Array{Float64,2}
```

kleinschmidt:
I'm closing this in favor of using `0` for the LHS, which should be fully supported with `hasresponse` once #144 lands.

PR #144, opened 2019-08-28
Fix intercept detector traits and hasresponse

kleinschmidt:
This improves support for using `0` on the LHS for a one-sided formula by making
sure that the `hasintercept` and `omitsintercept` traits are detected correctly
when `0` is present on the LHS.  Currently `0` on the LHS triggers the "omits
intercept" trait which blocks the default inclusion of the intercept for
`StatisticalModel` contexts.

This also updates the `hasresponse` detector to treat LHS `0` as no response.

### DIFF
--- a/src/terms.jl
+++ b/src/terms.jl
@@ -559,14 +559,19 @@ StatsBase.coefnames(t::InteractionTerm) =
 ################################################################################
 # old Terms features:
 
+hasintercept(f::FormulaTerm) = hasintercept(f.rhs)
 hasintercept(t::AbstractTerm) = InterceptTerm{true}() ∈ terms(t) || ConstantTerm(1) ∈ terms(t)
+omitsintercept(f::FormulaTerm) = omitsintercept(f.rhs)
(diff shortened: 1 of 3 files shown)


src/terms.jl
```diff
@@ -559,14 +559,19 @@ StatsBase.coefnames(t::InteractionTerm) =
...
     InterceptTerm{false}() ∈ terms(t) ||
     ConstantTerm(0) ∈ terms(t) ||
     ConstantTerm(-1) ∈ terms(t)
 
 hasresponse(t) = false
-hasresponse(t::FormulaTerm{LHS}) where {LHS} = LHS !== nothing
+hasresponse(t::FormulaTerm) =
+    t.lhs !== nothing && 
+    t.lhs != ConstantTerm(0) &&
+    t.lhs != InterceptTerm{false}
```
oxinabox: These could all be `!==` I think.

Though the last is missing it's `()`

kleinschmidt: Indeed.

Issue #141, opened 2019-08-21
Performances 0.5 vs 0.6


matthieugomez:
On my computer, StatsModels  `v0.6.0` is much slower than `v0.5.0` to construct a `ModelMatrix`. 

```
N = 10000000
df = DataFrame(y = randn(N), x1 = randn(N), x2 = randn(N))
@time ModelMatrix(ModelFrame(@formula(y~x1 + x2), df))
#> 0.365609 seconds (930 allocations: 840.530 MiB, 12.85% gc time)
```
and in v0.6
```
N = 10000000
df = DataFrame(y = randn(N), x1 = randn(N), x2 = randn(N))
@time ModelMatrix(ModelFrame(@formula(y~x1 + x2), df))
#>  0.838185 seconds (278 allocations: 764.150 MiB, 43.21% gc time)
```

kleinschmidt:
Interesting, I can't reproduce this on my laptop (I get about 1.2 for each of those if I use a `NamedTuple` instead of a DataFrame for 0.6, or ~1.3-4 (depending on the run) if I use a DataFrame).  I thought you might be getting some compile time but the first run is actually even faster for 0.6 so I'm not sure what's driving this.

That being said, I don't doubt that there's a possible performance hit with 0.6.  Could you try profiling to see where the slowdown is happening?  And has performance become a problem in real use-cases for you?

matthieugomez:
Environment with the best performances
```
(Try) pkg> status
    Status `~/FormerDataFrames/Project.toml`
  [a93c6f00] DataFrames v0.17.1 ⚲
  [3eaba693] StatsModels v0.5.0 ⚲
```
Note that, contrary to what I said in my previous posts, I actually use both an old version of DataFrames and an old version of StatsModels. Using either the current version of DataFrames or the current version of StatsModels seems to give worse performances.
Profiling is here:
https://gist.github.com/matthieugomez/5a38fa67716c21117466d1777fd60a93

My current Environment has:
```
  [3eaba693] StatsModels v0.6.3
  [a93c6f00] DataFrames v0.19.2
```
Profiling is here:
https://gist.github.com/matthieugomez/4025512d58ff1798ddda9c976d019ecc

It is not a big deal, but it means we're now slower than Stata for simple linear models. For now, Stata does regression of `y` on `x1` and `x2` under 1s, vs 1.2s in GLM (on my computer).

kleinschmidt:
[...]

PR #153, opened 2019-09-24
[WIP] Handle missing values 

palday:
When done, this will address #141.

Also todo: handle missing in categorical variables. Currently, missing is treated as an additional level, which doesn't seem like the correct thing.

### DIFF
--- a/src/schema.jl
+++ b/src/schema.jl
@@ -60,7 +60,7 @@ Base.haskey(schema::Schema, key) = haskey(schema.schema, key)
 Compute all the invariants necessary to fit a model with `terms`.  A schema is a dict that
 maps `Term`s to their concrete instantiations (either `CategoricalTerm`s or
 `ContinuousTerm`s.  "Hints" may optionally be supplied in the form of a `Dict` mapping term
-names (as `Symbol`s) to term or contrast types.  If a hint is not provided for a variable, 
+names (as `Symbol`s) to term or contrast types.  If a hint is not provided for a variable,
 the appropriate term type will be guessed based on the data type from the data column: any
 numeric data is assumed to be continuous, and any non-numeric data is assumed to be
 categorical.
@@ -91,7 +91,7 @@ StatsModels.Schema with 1 entry:
   y => y
 ```
 
-Note that concrete `ContinuousTerm` and `CategoricalTerm` and un-typed `Term`s print the 
+Note that concrete `ContinuousTerm` and `CategoricalTerm` and un-typed `Term`s print the
 same in a container, but when printed alone are different:
 
 ```jldoctest 1
@@ -176,6 +176,8 @@ concrete_term(t::Term, d) = concrete_term(t, d, nothing)
 concrete_term(t, d, hint) = t
 
 concrete_term(t::Term, xs::AbstractVector{<:Number}, ::Nothing) = concrete_term(t, xs, ContinuousTerm)
+# and for missing values
+concrete_term(t::Term, xs::AbstractVector{Union{Missing,T}} where T<:Number, ::Nothing) = concrete_term(t, xs, ContinuousTerm)
 function concrete_term(t::Term, xs::AbstractVector, ::Type{ContinuousTerm})
     μ, σ2 = StatsBase.mean_and_var(xs)
     min, max = extrema(xs)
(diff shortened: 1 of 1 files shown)


kleinschmidt:
This generally looks good but needs tests.  I think the problem I ran into was that `copy(::Missing)` isn't defined, but `modelcols(::ContinuousTerm, ...)` uses `copy` on teh elements of the underlying vector, instead of just writing the elements into a new vector.  The idea was that maybe someone has a custom type that is mutable but they define it as continuous (for instance, Survival.jl uses a custom `EventTime` struct, which isn't mutable but still it's not a stretch), but maybe that's being too defensive.

palday:
Yeah, that's about as far as I had gotten as well when I ran out of steam late local time. `deepcopy(::Missing)` is defined, and given that `missing` is a singleton, having `copy(m::Missing) = deepcopy(m)` seems like a reasonable definition, but probably one that doesn't belong here but rather in `Base`.

src/terms.jl
```diff
@@ -483,6 +483,13 @@ modelcols(t::Term, d::NamedTuple) =
 modelcols(ft::FunctionTerm{Fo,Fa,Names}, d::NamedTuple) where {Fo,Fa,Names} =
     ft.fanon.(getfield.(Ref(d), Names)...)
 
+# this is weird, but using import Base: copy leads to exporting type piracy
+# for non missing values, the compiler should hopefully optimize down the extra
+# layer of indirection
+function copy end
+copy(x::Any) = Base.copy(x)
+copy(m::Missing) = m
```
kleinschmidt: I think this feels like a bad idea.  Maybe better to just drop the use of `copy` (and broadcast `identity` or use `copy` on the vector).

palday:
Ignore the horrible hacking of `getindex` -- that will not stay.

src/schema.jl
```diff
@@ -176,6 +176,8 @@ concrete_term(t::Term, d) = concrete_term(t, d, nothing)
 concrete_term(t, d, hint) = t
 
 concrete_term(t::Term, xs::AbstractVector{<:Number}, ::Nothing) = concrete_term(t, xs, ContinuousTerm)
+# and for missing values
+concrete_term(t::Term, xs::AbstractVector{Union{Missing,T}} where T<:Number, ::Nothing) = concrete_term(t, xs, ContinuousTerm)
```
Nosferican: Couldn't those lines be
```
concrete_term(t::Term, xs::AbstractVector{<:Union{<:Number,<:Union{Missing,<:Number}}}) =
    concrete_term(t, xs, ContinuousTerm)
```
I don't think we need/should specialize.

kleinschmidt: Wouldn't just having `AbstractVector{<:Union{Missing, <:Number}}` work?

```
julia> [1,2] isa AbstractVector{<:Union{Missing,<:Number}}
true

julia> [1,2, missing] isa AbstractVector{<:Union{Missing,<:Number}}
true
```

greimel:
It would be nice if the following

```julia
julia> using DataFrames, StatsModels

julia> df = DataFrame(t=1:5, x=(1:5).*2.0, y=1:5.0, z=[12; missing; 16.:2:20])

julia> f = @formula(y ~ x + lag(x) + z + lag(z))
```

gave this outcome
```julia
julia> df_hand0 = transform(df, :z => lag => :z_lagged, :x => lag => :x_lagged)
5×6 DataFrame
 Row │ t      x        y        z          z_lagged   x_lagged  
     │ Int64  Float64  Float64  Float64?   Float64?   Float64?  
─────┼──────────────────────────────────────────────────────────
   1 │     1      2.0      1.0       12.0  missing    missing   
   2 │     2      4.0      2.0  missing         12.0        2.0
   3 │     3      6.0      3.0       16.0  missing          4.0
   4 │     4      8.0      4.0       18.0       16.0        6.0
   5 │     5     10.0      5.0       20.0       18.0        8.0

julia> df_hand = df_hand0[completecases(df_hand0),:] |> disallowmissing!

julia> f_hand = @formula(y ~ x + x_lagged + z + z_lagged)

julia> ff_hand = apply_schema(f_hand, schema(f_hand, df_hand))

julia> modelmatrix(ff_hand, df_hand)
2×4 Matrix{Float64}:
  8.0  6.0  18.0  16.0
 10.0  8.0  20.0  18.0
```

(If you happen to make this work in this PR, feel free to turn this code into a test.)

PR #155, opened 2019-09-28
Document support for special terms at run-time, one-sided formula

kleinschmidt:
This adds some notes in the documentation for package developers about supporting run-time use of special terms, expanding on a short comment from slack with @Nosferican.  It also updates the `poly`/`PolyTerm` example to follow the revised guidelines.


docs/src/internals.md
```diff
@@ -401,23 +403,29 @@ poly(x, n) = x^n
...
+struct PolyTerm{T,D} <: AbstractTerm
+    term::T
+    deg::D
 end
 
 Base.show(io::IO, p::PolyTerm) = print(io, "poly($(p.term), $(p.deg))")
 
-function StatsModels.apply_schema(t::FunctionTerm{typeof(poly)},
+# first pass: create PolyTerm
+StatsModels.apply_schema(t::FunctionTerm{typeof(poly)},
```
oxinabox: Because this is split onto so many lines.
I think it would be much easier to read if it stated with `function`
and ended with `end`

docs/src/internals.md
```diff
@@ -401,23 +403,29 @@ poly(x, n) = x^n
...
+                         sch::StatsModels.Schema,
+                         Mod::Type{<:POLY_CONTEXT}) =
+    apply_schema(PolyTerm(t.args_parsed...), sch, Mod)
+
+# second pass: apply schema to internal Terms and check for proper types
```
oxinabox: Why do we now do this in two passes?

docs/src/internals.md
```diff
@@ -568,40 +578,110 @@ a formula with a call to `poly`:
...
+b^1          2.13096    0.100552   21.1926    <1e-98   1.93388     2.32804
+b^2          3.1132     0.0813107  38.2877    <1e-99   2.95384     3.27257
+──────────────────────────────────────────────────────────────────────────
 
 ```
 
 (a `GeneralizeLinearModel` with a `Normal` distribution is equivalent to a
 `LinearModel`)
 
+### [Making special syntax "runtime friendly"] (@id extend-runtime)
```
oxinabox: Should this section be above the last?
It seems more closely re.ated to the example

docs/src/internals.md
kleinschmidt: I was gonna say, keep reading ;)

Yes a pointer would be appropriate

docs/src/internals.md
kleinschmidt: Everything in this section is about the same example, and the thing about the context is really important (more so than supporting run-time creation I'd say).  Then again, I'm not totally committed to this order.  I'd just written the earlier parts in one go and didn't want to have to edit those sections but I can have another look.

docs/src/internals.md
```diff
@@ -485,20 +502,158 @@ julia> coefnames(f.rhs)
...
+
+When used from the `@formula` macro, special syntax relies on dispatching on the
+`FunctionTerm{MyFunction}` type.  But when creating a formula at runtime
+without the `@formula` macro, `FunctionTerm`s aren't available, and so care must
+be taken to make sure you provide a runtime replacement.  The example for `poly`
+above shows how to do this, but we spell it out here in more detail.
+
+The first step is to make sure you can create a schema-less instance of the
+`AbstractTerm` that implements your special syntax behavior.  That's why for the
```
nalimilan: One of "that's why" and "that means" should be dropped.

docs/src/internals.md
```diff
@@ -485,20 +502,158 @@ julia> coefnames(f.rhs)
...
+julia> modelcols(ff.rhs, data)
+4×4 Array{Float64,2}:
+ 1.0  0.488613  1.0   1.0
+ 1.0  0.210968  2.0   4.0
+ 1.0  0.951916  3.0   9.0
+ 1.0  0.999905  4.0  16.0
+```
+
+This is workable if a little verbose, so the third (optional) step is to provide
+additional "runtime" methods for the function that's used in the `@formula`
```
nalimilan:
```suggestion
additional "runtime" methods for the function that are used in the `@formula`
```

docs/src/formula.md
```diff
@@ -279,6 +284,20 @@ Predictors:
...
+    Even though the `@formula` macro supports arbitrary julia functions,
+    runtime (programmatic) formula construction does not.  This is because to
+    resolve a symbol giving a function's _name_ into the actual _function_
+    itself, it's necessary to `eval`.  In practice this is not often an issue,
+    _except_ in cases where a package provides special syntax by overloading a
+    function (like `|` for
+    [MixedModels.jl](https://github.com/dmbates/MixedModels.jl), or `absorb`
+    for [Econometrics.jl](https://github.com/Nosferican/Econometrics.jl)).  In
+    these cases, you should use the corresponding constructors for the actual
+    terms themselves (e.g., `RanefTerm` and `FixedEffectsTerm` respectively), as
```
matthieugomez: How is the user supposed to discover these names?

It seems to me that a simpler alternative would be to ask package managers to directly overload `poly` `|` and `absorb` for terms, so that users can do `absorb(Terms(:a))`, `poly(Term(:a), 2)` etc. This also limits the namespace pollution (I guess I may also want to define a FixedEffectTerm in the FixedEffectModels package).

docs/src/internals.md
kleinschmidt: "function" is the head of this relative clause so single number is appropriate no?  That is, it's the function that's used, not the methods...although now I see that this is potentially structurally ambiguous :)

docs/src/formula.md
kleinschmidt: Keep reading :)  I suppose it would be possible to include that in the main part of the example, too.

kleinschmidt: (or rather, instead of having it separately down below)

kleinschmidt: I've decided you're right and revised the whole section accordingly (including a note to make sure you export `poly`).  have another read?

matthieugomez: This looks great. Thanks!

PR #157, opened 2019-10-02
get rid of CategoricalArrays

kleinschmidt:
Depends on having a levels method defined in DataAPI (or elsewhere), ~~which it currently does not~~ which it does since 1.1.0.  Also, there were a few functions from Missings that snuck in because CategoricalArrays reexports it.


kleinschmidt:
see JuliaData/DataAPI.jl#12

kleinschmidt:
This will also depend on Categorical Arrays pulling in `levels` from DataAPI instead of Missing

kleinschmidt:
@nalimilan pointed out that we can still use the specialized method if we use `DataAPI.refarray`

src/modelframe.jl
```diff
@@ -53,11 +53,12 @@ function _nonmissing!(res, col)
     end
 end
 
-function _nonmissing!(res, col::CategoricalArray{>: Missing})
-    @inbounds for (i, el) in enumerate(col.refs)
-        res[i] &= el > 0
-    end
-end
+# TODO: find another way to optimize this without taking a dependency
```
nalimilan: Just remove this thanks to https://github.com/JuliaData/CategoricalArrays.jl/pull/218.

kleinschmidt: does this mean that we need to change these methods to use broadcast instead of doing a loop?

kleinschmidt:
I think this is ready now that Missings, DataAPI, and CategoricalArrays have all had releases.

Issue #120, opened 2019-06-21
enhanced contrast coding features


kleinschmidt:
This is a "meta issue" to keep track of features that would be "nice to have" for contrast coding.  I'll update this as other things come up.

# Additional contrast types for ordinal variables

It would be useful to support some additional contrast types that are useful for ordinal variables (where the levels have a well defined order)

[ ] - Repeated/successive differences (each contrast tests for difference between two adjacent levels; `contr.sdiff` from MASS)
[ ] - Polynomial contrasts (orthogonal polynomials that test for linear, quadratic, etc. trends in the ordered levels).
[ ] - Check correctness of helmert coding (I have a nagging suspicion that this might not be quite correctly scaled or something)

# Interpretation of coefficients

The (generalized) inverse of the contrasts matrix gives the loading of each condition's mean on each coefficient, which is useful for interpretation of the coefficients in terms of the underlying cells.  We should provide an interface to extract this, including sensitivity to when an intercept column or other contrasts need to be included (e.g., non-centered contrasts or the presence of interactions with other contrast-coded variables).  We should also provide an interface to work with this for users, to e.g. interpret the meaning of the intercept coefficient in the presence of different contrast coding schemes.

oxinabox:
We should have an `OrdinalTerm` type to go with that

PR #158, opened 2019-10-02
Implement HypothesisCoding and ~un-export~ deprecate ContrastsCoding

kleinschmidt:
`ContrastsCoding` is a footgun unless you really know what you're doing.
Hypothesis coding is a more natural way of specifying manual contrasts, since it
directly expresses differences between conditional means of levels in the matrix
(which is the pseudo-inverse of the contrasts matrix).  This PR introduces a
`HypothesisCoding <: AbstractContrasts` which is exported, and un-exports
`ContrastsCoding` (and adds a note in the docs/help text that points to
hypotheses)

### DIFF
--- a/src/contrasts.jl
+++ b/src/contrasts.jl
@@ -368,10 +368,83 @@ function contrasts_matrix(C::HelmertCoding, baseind, n)
 end
 
 """
-    ContrastsCoding(mat::Matrix[, base[, levels]])
+    HypothesisCoding(hypotheses::Matrix[, levels]])
+
+Specify how to code a categorical variable in terms of a *hypothesis matrix*.
+For a variable with ``k`` levels, this should be a ``k-1 \times k`` matrix.
+Each row of the hypothesis corresponds to the hypothesis tested by the
+corresponding predictor, given as the weights given to the "cell mean" of each
+of the ``k`` values of the predictor.
+
+For instance, if we have a variable which has four levels A, B, C, and D, and we
+want to test the hypothesis that the difference between the average outcome for
+levels A and B is different from zero, the corresponding row of the hypothesis
+matrix would be `[-1, 1, 0, 0]`.  Likewise, to test whether the difference
+between B and C is different from zero, the hypothesis vector would be `[0, -1,
+1, 0]`.  To test each "successive difference" hypothesis, the full hypothesis
+matrix would be
+
+```jldoctest hyp
+julia> sdiff_hypothesis = [-1  1  0  0
+                            0 -1  1  0
+                            0  0 -1  1];
+```
+
+Contrasts are derived the hypothesis matrix by taking the pseudoinverse:
+
+```jldoctest hyp
+julia> sdiff_contrasts = pinv(sdiff_hypothesis)
+4×3 Array{Float64,2}:
+ -0.75  -0.5  -0.25
+  0.25  -0.5  -0.25
+  0.25   0.5  -0.25
+  0.25   0.5   0.75
+```
+
+Which is what is produced by constructing a [`ContrastsMatrix`](@ref) from a
+`HypothesisCoding` instance:
+
+```jldoctest hyp
+julia> StatsModels.ContrastsMatrix(HypothesisCoding(sdiff_hypothesis), ["a", "b", "c", "d"]).matrix
+4×3 Array{Float64,2}:
+ -0.75  -0.5  -0.25
+  0.25  -0.5  -0.25
+  0.25   0.5  -0.25
+  0.25   0.5   0.75
+```
+
+
+"""
+mutable struct HypothesisCoding <: AbstractContrasts
+    hypotheses::Matrix
+    contrasts::Matrix
+    base::Nothing
+    levels::Union{Vector,Nothing}
+
+    function HypothesisCoding(hypotheses, base, levels)
+        contrasts = pinv(hypotheses)
+        check_contrasts_size(contrasts, levels)
+        new(hypotheses, contrasts, base, levels)
+    end
+end
+
+HypothesisCoding(mat::Matrix; levels=nothing) =
+    HypothesisCoding(mat, nothing, levels)
+
+function contrasts_matrix(C::HypothesisCoding, baseind, n)
+    check_contrasts_size(C.contrasts, n)
+    C.contrasts
+end
+
+
+"""
+    StatsModels.ContrastsCoding(mat::Matrix[, base[, levels]])
 
 Coding by manual specification of contrasts matrix. For k levels, the contrasts
-must be a k by k-1 Matrix.
+must be a k by k-1 Matrix.  The contrasts in this matrix will be copied directly
+into the model matrix; if you want to specify your contrasts as hypotheses (i.e., 
+weights assigned to each group's cell mean), you should use 
+[`HypothesisCoding`](@ref) instead.
 """
 mutable struct ContrastsCoding <: AbstractContrasts
     mat::Matrix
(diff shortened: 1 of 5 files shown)


kleinschmidt:
I've also implemented sequential difference/sliding/repeated contrasts coding for the hell of it (see #120)

src/contrasts.jl
```diff
@@ -368,25 +368,144 @@ function contrasts_matrix(C::HelmertCoding, baseind, n)
...
+    for col in 1:n-1
+        mat[1:col, col] .= col-n
+        mat[col+1:end, col] .= col
+    end
+    return mat ./ n
+end
+
+
+"""
+    HypothesisCoding(hypotheses::Matrix[, levels]])
```
nalimilan:
```suggestion
    HypothesisCoding(hypotheses::Matrix[, levels])
```

src/contrasts.jl
```diff
@@ -368,25 +368,144 @@ function contrasts_matrix(C::HelmertCoding, baseind, n)
 end
 
 """
-    ContrastsCoding(mat::Matrix[, base[, levels]])
+    SeqDiffCoding([base[, levels]])
+
+Codes each level in order to test "sequential difference" hypotheses, which
```
nalimilan:
```suggestion
Code each level in order to test "sequential difference" hypotheses, which
```

src/contrasts.jl
```diff
@@ -368,25 +368,144 @@ function contrasts_matrix(C::HelmertCoding, baseind, n)
...
+    return mat ./ n
+end
+
+
+"""
+    HypothesisCoding(hypotheses::Matrix[, levels]])
+
+Specify how to code a categorical variable in terms of a *hypothesis matrix*.
+For a variable with ``k`` levels, this should be a ``k-1 \times k`` matrix.
+Each row of the hypothesis corresponds to the hypothesis tested by the
```
nalimilan:
```suggestion
Each row of the matrix corresponds to the hypothesis tested by the
```

src/contrasts.jl
```diff
@@ -368,25 +368,144 @@ function contrasts_matrix(C::HelmertCoding, baseind, n)
...
+```jldoctest hyp
+julia> sdiff_hypothesis = [-1  1  0  0
+                            0 -1  1  0
+                            0  0 -1  1];
+```
+
+Contrasts are derived the hypothesis matrix by taking the pseudoinverse:
+
+```jldoctest hyp
+julia> sdiff_contrasts = pinv(sdiff_hypothesis)
```
nalimilan:
```suggestion
julia> pinv(sdiff_hypothesis)
```

src/contrasts.jl
```diff
@@ -368,25 +368,144 @@ function contrasts_matrix(C::HelmertCoding, baseind, n)
...
+4×3 Array{Float64,2}:
+ -0.75  -0.5  -0.25
+  0.25  -0.5  -0.25
+  0.25   0.5  -0.25
+  0.25   0.5   0.75
+```
+
+Which is what is produced by constructing a [`ContrastsMatrix`](@ref) from a
```
nalimilan:
```suggestion
The above matrix is what is produced by constructing a [`ContrastsMatrix`](@ref) from a
```

src/contrasts.jl
```diff
@@ -368,25 +368,144 @@ function contrasts_matrix(C::HelmertCoding, baseind, n)
...
+ -0.75  -0.5  -0.25
+  0.25  -0.5  -0.25
+  0.25   0.5  -0.25
+  0.25   0.5   0.75
+```
+
+
+"""
+mutable struct HypothesisCoding <: AbstractContrasts
+    hypotheses::Matrix
```
nalimilan: Shouldn't this be parameterized on the eltype?

src/contrasts.jl
```diff
@@ -368,25 +368,144 @@ function contrasts_matrix(C::HelmertCoding, baseind, n)
...
+    contrasts::Matrix
+    base::Nothing
+    levels::Union{Vector,Nothing}
+
+    function HypothesisCoding(hypotheses, base, levels)
```
nalimilan: Why specify `base` if it's always `nothing`?

src/contrasts.jl
```diff
@@ -368,25 +368,144 @@ function contrasts_matrix(C::HelmertCoding, baseind, n)
...
+end
+
+
+"""
+    HypothesisCoding(hypotheses::Matrix[, levels]])
+
+Specify how to code a categorical variable in terms of a *hypothesis matrix*.
+For a variable with ``k`` levels, this should be a ``k-1 \times k`` matrix.
+Each row of the hypothesis corresponds to the hypothesis tested by the
+corresponding predictor, given as the weights given to the "cell mean" of each
```
nalimilan: I don't really understand that phrasing. The repetition of "given" doesn't help. :-)

src/contrasts.jl
kleinschmidt: This isn't used anywhere but it is more to help with reading...

src/contrasts.jl
kleinschmidt: I think there's code that tries to access the field.  I could overload getproperty for this type instead though, or refactor that code...

src/contrasts.jl
kleinschmidt: Maybe, although it's going to get converted to whatever the eltype of the final model matrix is anyway.  But maybe it would help the compiler?

src/contrasts.jl
kleinschmidt: how about "expressed as the weights of the "cell mean" of each of the ``k`` values of the predictor"?

kleinschmidt:
For non-breaking-ness, perhaps it would be better to deprecate ContrastsCoding with a message that it will be un-exported in the next minor release?  Rather than removing it right now.

src/contrasts.jl
nalimilan: Since `HypothesisCoding` is new, how could existing code try to access that field? :-)

 Anyway, in general I think fields are considered private unless specified otherwise. And unexporting `ContrastsCoding` is much more breaking. BTW, maybe we should keep exporting it for now until we want to make another breaking release? It's simpler to break many things at the same time than force people to update small things continuously.

src/contrasts.jl
nalimilan: Yeah, if the compiler can know the concrete type of such objects in typical situations (which I think is the case), it will generate much more efficient code. That should matter especially if you iterate over rows.

Though I've noticed this affects other contrasts, so maybe for a separate PR.

src/contrasts.jl
```diff
@@ -368,25 +368,144 @@ function contrasts_matrix(C::HelmertCoding, baseind, n)
...
+of the ``k`` values of the predictor.
+
+For instance, if we have a variable which has four levels A, B, C, and D, and we
+want to test the hypothesis that the difference between the average outcome for
```
nalimilan:
```suggestion
want to test the hypothesis that the difference between the average outcomes for
```

src/contrasts.jl
nalimilan: Yeah... I guess I don't really understand what "cell means" refers to in this context, but maybe that's OK.

nalimilan:
> For non-breaking-ness, perhaps it would be better to deprecate ContrastsCoding with a message that it will be un-exported in the next minor release? Rather than removing it right now.

Yes, that sounds better, but I always forget whether that's possible or not. We could also just keep exporting it for now, and only remove it in the next minor release (when we want to break other, more fundamental things).

src/contrasts.jl
kleinschmidt: Sounds good.  Let's keep it exported but maybe add a warning to teh constructor or something?

src/contrasts.jl
kleinschmidt: I think it only affect this and `ContrastsCoding` (the only other one that involves a manually specified matrix)

src/contrasts.jl
kleinschmidt: I added a depwarn because the deprecations weren't quite doing it correctly (giving a spurious method overwritten warning)

src/contrasts.jl
kleinschmidt: And anyway, I don't think the contrasts themselves ever get used in any hot code paths (that's all the `ContrastsMatrix` struct...)

src/contrasts.jl
kleinschmidt: The base field is used in the constructor of `ContrastsMatrix`.  Could change that to use an accessor method but that seems like overkill at this point.

src/contrasts.jl
```diff
@@ -414,16 +414,18 @@ end
...
 """
-    HypothesisCoding(hypotheses::Matrix[, levels]])
+    HypothesisCoding(hypotheses::Matrix[, levels])
 
 Specify how to code a categorical variable in terms of a *hypothesis matrix*.
 For a variable with ``k`` levels, this should be a ``k-1 \times k`` matrix.
-Each row of the hypothesis corresponds to the hypothesis tested by the
-corresponding predictor, given as the weights given to the "cell mean" of each
-of the ``k`` values of the predictor.
+Each row of the matrix corresponds corresponds to a hypothesis about the mean
```
nalimilan:
```suggestion
Each row of the matrix corresponds to a hypothesis about the mean
```

PR #160, opened 2019-10-04
Fix predict with confidence interval

nalimilan:
DataFrames aren't a dependency anymore.

Fixes https://github.com/JuliaStats/StatsModels.jl/issues/159. Covered by tests at https://github.com/JuliaStats/GLM.jl/pull/335 (we should add tests with a dummy model type at some point).

### DIFF
--- a/src/statsmodel.jl
+++ b/src/statsmodel.jl
@@ -131,26 +131,26 @@ StatsBase.adjr2(mm::TableRegressionModel) = adjr2(mm.model)
 StatsBase.r2(mm::TableRegressionModel, variant::Symbol) = r2(mm.model, variant)
 StatsBase.adjr2(mm::TableRegressionModel, variant::Symbol) = adjr2(mm.model, variant)
 
-function _return_predictions(yp::AbstractVector, nonmissings, len)
+function _return_predictions(T, yp::AbstractVector, nonmissings, len)
     out = missings(eltype(yp), len)
     out[nonmissings] = yp
     out
 end
 
-function _return_predictions(yp::AbstractMatrix, nonmissings, len)
+function _return_predictions(T, yp::AbstractMatrix, nonmissings, len)
     out = missings(eltype(yp), (len, 3))
     out[nonmissings, :] = yp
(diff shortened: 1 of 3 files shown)


src/statsmodel.jl
```diff
@@ -131,26 +131,26 @@ StatsBase.adjr2(mm::TableRegressionModel) = adjr2(mm.model)
 StatsBase.r2(mm::TableRegressionModel, variant::Symbol) = r2(mm.model, variant)
 StatsBase.adjr2(mm::TableRegressionModel, variant::Symbol) = adjr2(mm.model, variant)
 
-function _return_predictions(yp::AbstractVector, nonmissings, len)
+function _return_predictions(T, yp::AbstractVector, nonmissings, len)
     out = missings(eltype(yp), len)
     out[nonmissings] = yp
     out
```
nalimilan: Note that in this case, we return a mere vector (not a data frame) even if the input is a data frame. I guess that's the most logical output.

kleinschmidt: I think that's what the consensus was right?  I remember their being some extensive discussion about this...

kleinschmidt:
Weird how this snuck in; I guess the predict PRs were opened while #71 was still being finalized and didn't touch any of the same code so I didn't notice.  We really ought to have tests for this though, could you add some?

src/statsmodel.jl
nalimilan: Ah yes that was https://github.com/JuliaStats/StatsModels.jl/pull/81.

nalimilan:
> Weird how this snuck in; I guess the predict PRs were opened while #71 was still being finalized and didn't touch any of the same code so I didn't notice.

Yes that seems to be what happened with https://github.com/JuliaStats/StatsModels.jl/pull/81.

> We really ought to have tests for this though, could you add some?

 JuliaStats/GLM.jl#335 adds tests in GLM. I'll have a look at adding tests here.

nalimilan:
OK, tests added!

kleinschmidt:
Oh and bump the patch version?

nalimilan:
Done.

PR #162, opened 2019-10-17
Add lrtest

nalimilan:
Heavily Inspired by `ftest` in GLM. Also introduce an `isnested` function which can be overloaded by modeling packages to protect users from comparing non-nested models: a warning is printed if no method has been defined by a model type; if it's defined, an error is thrown for non-nested models.


kleinschmidt:
Is it wise to take a dependency on Distributions?  That's rather heavy and I thought also pulls in RMath too, right?

nalimilan:
Yeah that's not great (at least the Rmath part). But I don't really see where this could live. I guess we could put this in a separate package which would be loaded by StatsKit, but a package with a single function is a bit funny... HypothesisTests would be a possible home, but currently it doesn't have anything related to models.

oxinabox:
Pros and cons of making a package for just this?
Maybe ModellingTests.jl or some such?

nalimilan:
The pro is not adding deps to StatsModels. The cons are that it's annoying to have to load a package just for one function, it's harder to discover and it's more work to maintain.

kleinschmidt:
Maybe another thing that belongs in StatsModelsExtras.jl?

nalimilan:
"Another" in addition to what? Indeed if we have several things to put there, that would justify the creation of a package.

kleinschmidt:
There was some discussion of putting more "exotic" custom syntax there (like poly regression, splines, maybe the lead/lag stuff).

oxinabox:
> There was some discussion of putting more "exotic" custom syntax there (like poly regression, splines, maybe the lead/lag stuff).

This feels different to that,
but still from a dependency management point it would be a place to start

hmatuschek:
Concerning adding dependencies: I just implemented the complementary prob. func. of the chi-squared dist. assuming integer degrees of freedom (https://gist.github.com/hmatuschek/17d2cba4ca15273aab9ad2647500c47b). This might be an option until the SpecialFunctions package provides a proper impl. of the incomplete gamma functions.

Nosferican:
What's the status of this?

nalimilan:
I think we need to create StatsModelsExtras, or alternatively create StatsModelsBase and move existing features to it.

Another solution would be to use @hmatuschek's code for the Chi squared distribution, but it would be annoying to duplicate code just to avoid depending on Distributions.jl (and we may need other things in the future).

andreasnoack:
I think it would be good, if nothing else, to have the name defined somewhere upstream such that packages can extend the same `lrtest` function. I think having the name somewhere is more important than a common interface since the function might need some customization for each application.

nalimilan:
Actually I don't think packages will often have to specialize this function: it depends only on very generic functions like `loglikelihood`, `dof` and the newly added `isnested`. It would be too bad if they ended up reimplementing it just because we haven't yet split StatsModels into a low-level and a high-level package.

andreasnoack:
I’m not against having an implementation here. I agree that a basic version should be quite unambiguously defined from the components you mention. I just think it’s important to get the name defined upstream right away and I also don’t think we should try to insist on a return type. I.e. if a downstream package has reasons to define this slightly differently but still overload the same function, we should consider bad.

palday:
There may be a middle ground: returning the p-value as `nothing` until a way around the `Distributions` dependency is found. There's a bit more fiddling to be done so that downstream packages that already `depend` on `Distributions` can get the p-value, but at least the function name and return type are available.

nalimilan:
I think we should just settle the dependencies issue by deciding whether we want StatsModelsBase+StatsModels or StatsModels+StatsModelsExtras. Given that there's going to be DataFramesBase, I suggest we go the former route for consistency in names.

Then I think the idea was to move modeling interfaces from StatsBase to StatsModels/StatsModelsBase.

andreasnoack:
So I just checked and among the 24 packages that depend directly on StatsModels only Microeconometrics (surprisingly!) and RMath (not sure why it depends on StatsModels) don't depend on Distributions so I think we should consider adding Distributions as a dependency here since it most likely won't have any implications for any users.

nalimilan:
Why not. Though we had also been talking about adding more dependencies (e.g. for splines support), so we'd better decide right now whether they would be OK too. One advantage choosing the StatsModelsBase+StatsModels solution is that the split can be done at any time without breaking as long as the former reexports the latter: packages that want to get rid of too heavy dependencies we may want to add to StatsModels can switch to importing StatsModelsBase instead.

kleinschmidt:
Here's a proposal:

* We agree to split the package into `StatsModelsBase` and `StatsModels`.
* `lrtest` gets a stub definition in base (along with all the other modeling related names that are currently defined all over the place, esp. StatsBase)
* StatsModels takes Distributions as a dependency
* we merge this PR as is now and then do the split.

I'm less worried about packages that currently depend on StatsModels since StatsModels used to be a heavy dependency (depended on DataFrames); I'm more concerned with the packages that might consider taking a dependency on statsmodels in the future but not want to pull in something like distributions.  But having StatsModelsBase to provide methods for formulae seems like a good compromise to me.

nalimilan:
Sounds like a plan! I've rebased the PR, but CI fails on master currently due to DataFrames 0.21 (https://github.com/JuliaStats/StatsModels.jl/pull/179). Let me know if something else needs to be changed.

nalimilan:
CI passes now.

nalimilan:
@kleinschmidt I'll merge in a few days if you have no objections. :-)

kleinschmidt:
Did we already discuss bumping the version?

nalimilan:
I don't think so. Do you mean the patch or minor? (It's not breaking so I'd say patch, right?)

kleinschmidt:
it's not breaking but it DOES introduce different dependencies with potentially different compatibility bounds so it might end up messing with someone's installation?  I can never keep straight which things are sufficiently breaking to count as breaking.

Speaking of which, better add compatibility bounds for the extra dependencies or else automerge will fail.  Can compat helper do that?

nalimilan:
Adding more dependencies isn't supposed to break one's installation. The worst that can happen is that the package will stay at the previous release, which is exactly what will happen if we tag a breaking release (until dependencies have been updated). But as you prefer.

Good catch about bounds. I've added them for Distributions: after testing it turns out any version works (0.16 is the oldest version which supports Julia 1.0).

kleinschmidt:
Patch is fine by me, sorry for the slow response here.

PR #166, opened 2019-11-11
add upper bounds to please the automerge bots

kleinschmidt:


Project.toml
```diff
@@ -14,7 +14,10 @@ Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 [compat]
 julia = "1"
 DataAPI = "1.1"
-StatsBase = "≥0.22.0"
+DataStructures = "0.17.0"
+ShiftedArrays = "1.0.0"
+StatsBase = "0.22, 0.23, 0.24, 0.25, 0.26, 0.27, 0.28, 0.29, 0.30, 0.31, 0.32"
```
oxinabox: brutal

kleinschmidt: but just think what will happen to all the artisans making their living with hand-crafted bespoke compatibility ranges

Issue #111, opened 2019-06-13
coefnames should work on plain `Term`s


oxinabox:
before a formula has had a schema applied there is some kind of sensible fallback.

```
julia> coefnames(@formula(y~x))
ERROR: MethodError: no method matching coefnames(::Term)
```

PR #167, opened 2020-01-04
Indexable model

pdeffebach:
This is a very short PR that serves as a test for a feature that I think would be nice in StatsModels. It allows you to index a model by a `Term`. 

It is motivated by Stata functionality like ``_b[`var']`` which allows you to, get the `beta` coefficient for the column represented by `var`. This is really useful when making tables and graphs programatically. 

My approach is to take in a model and a `AbstractTerm`. Then check if the `AbstractTerm` matches, roughly, something in the model. If it does, match, it returns a `NamedTuple` with the coefficient name, coefficient, and the standard error.

```
julia> t = (y = rand(100), x = rand(100), b = rand(Bool, 100));
julia> m = lm(@formula(y ~ x + x & b), t);
julia> getparams(m, Term(:x)) 
```

Note that in the last line `Term(:x)` is not a `ContinuousTerm` or `CategoricalTerm`, I just match the `x.sym` paramter.

### DIFF
--- a/src/statsmodel.jl
+++ b/src/statsmodel.jl
@@ -196,3 +196,25 @@ function Base.show(io::IO, model::TableModels)
         end
     end
 end
+
+function getparams(model::TableModels, t)
+    m = model.mf.f.rhs.terms
+    i = 1
+    for x in m
+        if (t isa CategoricalTerm || t isa ContinuousTerm) &&
+           (x isa CategoricalTerm || x isa ContinuousTerm)
+            x.sym == t.sym && break
+        elseif t isa InteractionTerm && x isa InteractionTerm
+            xterms = map(s -> s.sym, x.terms)
+            tterms = map(s -> s.sym, t.terms)
+            xterms == tterms && break
+        elseif t isa InterceptTerm{true} && x isa InterceptTerm{true} 
+            break
+        else 
+            error("Not supported yet")
+        end
+        i += 1
+        
+    end
+    return (coefname = coefnames(model)[i], coef = coef(model)[i], stderror = stderror(model)[i])
+end
--- a/src/statsmodel.jl
+++ b/src/statsmodel.jl
@@ -197,24 +197,35 @@ function Base.show(io::IO, model::TableModels)
     end
 end
 
-function getparams(model::TableModels, t)
+"""
+    getparams(model::TableModels, t::AbstractTerm)
+
+Returns a named tuple of parameters corresponding to `t`. 
+
+Examples: 
+
+julia> t = (y = rand(100), x = rand(100), b = rand(Bool, 100));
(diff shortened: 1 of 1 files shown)


Tokazama:
I think this might begin to address a larger problem we have. Although we have the ability to go from a formula to a model, we don't have a way to do the reverse. I know this is difficult to ensure because it's up to packages like GLM.jl and MixedModels.jl to support this sort of behavior, but if we want to be able to plot the results of `lm(@formula(y ~ x + x & b), t)` then we need to be able to parse out the relevant terms as intercepts, slopes, etc.

kleinschmidt:
I like this idea generally, and it's related to other discussions about overhauling the modeling API (e.g., requiring that `StatisticalModel`s keep the formula themselves, instead of relying on the wrapper type like we currently do).  #32 I think is the relevant issue...

pdeffebach:
Presumably this would be a fall back, returning a Named Tuple for *just* the statistics models are required to have, if any, right? Most packages would have to write their own `getparams` function.

Tokazama:
After looking at #32 it seems like we want to move towards the formula interface and less so `TableModels`. Would it make sense to have a `getparams` for formulas so that it could be model agnostic? This wouldn't necessarily preclude the current PR, but would instead make it easier to extend to new types of terms.

Tokazama:
Related issue is https://github.com/JuliaStats/StatsModels.jl/issues/111, which would be extremely useful.

pdeffebach:
I'm reviving this. 

David do you have anything in mind for how exactly this should work? The inter-linking of PRs and Issues here suggests this problem is likely in a chicken-or-the-egg state. 

Is there concrete groundwork that really has to be done before we can implement this feature more?

palday:
What would this look like for `FormulaTerm`s? Random effects in MixedModels.jl are initially parsed as `FormulaTerm`s.

pdeffebach:
> What would this look like for `FormulaTerm`s? Random effects in MixedModels.jl are initially parsed as `FormulaTerm`s.

I don't know! Probably something like 

```
julia> getparams(m, Term(RandomEffect, :x: )) 
```

But my knowledge of all of this is pretty weak. 

@Tokazama I agree that #111 would be super useful here. It would make it a lot easier to index into a model.

pdeffebach:
Just realized that R's outputs are indexable via heavy use of named arrays

```
r$> confint(m_ols)                                                                                        
                 2.5 %    97.5 %
(Intercept)  3.8836484 5.5022774
exprop       0.3584959 0.6164465
latitude    -0.2919431 2.3197287

r$> confint(m_ols)["exprop", "2.5 %"]                                                                     
[1] 0.3584959
```

I regret having let this languish. Hopefully I can pick it up soon.

Issue #113, opened 2019-06-13
Should coeffnames be Symbols?


oxinabox:
It might be a bit breaking,
but `Symbol` is the type generally used for _The Name of a Thing_

and DataFrame column names are `Symbols`,
as are `ColumnTables`

Nosferican:
Another thing to consider is whether we want the type to change depending on the inputs,
```
data = [(x = 0,)]
coefnames(apply_schema(@formula(0 ~ 0), schema(data)))
coefnames(apply_schema(@formula(0 ~ 1), schema(data)))
coefnames(apply_schema(@formula(0 ~ 1 + x), schema(data)))
coefnames(apply_schema(@formula(1 ~ 0), schema(data)))
coefnames(apply_schema(@formula(1 ~ 1), schema(data)))
coefnames(apply_schema(@formula(1 ~ 1 + x), schema(data)))
```

Tokazama:
I really like the idea of using `Symbol` for coefnames instead of strings but what would be used for interaction terms? Would we just want to currently developed string name and convert it to a symbol?

oxinabox:
Sure, why not?

nalimilan:
As noted by @kleinschmidt at #169, it's not clear what would be the advantage of doing this. Yet it's quite disruptive. Can you develop?

Tokazama:
Just to  add to what @oxinabox said in the OP, terms are also using symbols as names so it seems very odd that the only place we are using strings to refer to variables is the coefficients.

The actually reason I started looking at this was because it's easier to speed up operations on symbols than strings. I was looking at ways to query coefficients within models and formulas.

oxinabox:
To repreat my original point: it puts us inline with how Tables.jl represents things.

Tokazama:
We should probably also address what we'd miss out on if we no longer used strings

nalimilan:
I see the point of consistency with Tables. OTOH, coefficients names are not exactly like terms, as you won't type them in a Julia-like syntax and they are often not valid Julia identifiers. This is clearly visible in the PR's tests: very often coefficient names cannot be typed nor printed directly using the `:` syntax. Compare:
```julia
julia> ["Intercept", "x1p: 6", "x1p: 7", "x1p: 8"]
4-element Array{String,1}:
 "Intercept"
 "x1p: 6"
 "x1p: 7"
 "x1p: 8"

[...]

PR #169, opened 2020-01-20
coefnames returns symbols

Tokazama:
This PR begins to address https://github.com/JuliaStats/StatsModels.jl/issues/113. Right now it errors on the `statsmodels` tests. It's due to a problem with `CoefTable`. I think I would need to make a PR to StatsBase giving `NoQuote` the ability to handle `Symbol`. Other than that it seems to pass all tests.

So if you like this PR I'll take the next step and see if I can fix the problem in StatsBase.


kleinschmidt:
Thanks for this! I don't think I'll have time to review this until mid Feb so ping me if I haven't weighed in by then.

mkborregaard:
This looks nice

src/temporal_terms.jl
```diff
@@ -54,7 +54,6 @@ function Base.show(io::IO, ll::LeadLagTerm{<:Any, F}) where F
     opname = string(nameof(F.instance))
     print(io, "$opname($(ll.term), $(ll.nsteps))")
 end
-function StatsBase.coefnames(ll::LeadLagTerm{<:Any, F}) where F
-    opname = string(nameof(F.instance))
-    coefnames(ll.term) .* "_$opname$(ll.nsteps)"
-end
+StatsBase.coefnames(ll::LeadLagTerm{<:Any, F}) where F = _llcoef(ll, coefnames(ll.term), string(nameof(F.instance)))
+_llcoef(ll::LeadLagTerm, t::Symbol, opname) = Symbol(t, "_$opname$(ll.nsteps)")
```
kleinschmidt: why can't you use the original broadcast based solution?

src/terms.jl
```diff
@@ -543,18 +543,48 @@ vectorize(x) = [x]
     coefnames(term::AbstractTerm)
 
 Return the name(s) of column(s) generated by a term.  Return value is either a
-`String` or an iterable of `String`s.
+`Symbol` or an iterable of `String`s.
```
kleinschmidt: missed a String

src/terms.jl
```diff
@@ -543,18 +543,48 @@ vectorize(x) = [x]
...
 """
+StatsBase.coefnames(t::Term) = t.sym
 StatsBase.coefnames(t::FormulaTerm) = (coefnames(t.lhs), coefnames(t.rhs))
-StatsBase.coefnames(::InterceptTerm{H}) where {H} = H ? "(Intercept)" : []
-StatsBase.coefnames(t::ContinuousTerm) = string(t.sym)
-StatsBase.coefnames(t::CategoricalTerm) = 
-    ["$(t.sym): $name" for name in t.contrasts.termnames]
-StatsBase.coefnames(t::FunctionTerm) = string(t.exorig)
-StatsBase.coefnames(ts::TupleTerm) = reduce(vcat, coefnames.(ts))
+StatsBase.coefnames(::InterceptTerm{H}) where {H} = H ? Symbol(:Intercept) : []  # this seems like the wrong thing to return
```
kleinschmidt: Can we keep the parens?  Then you'd have `Symbol("(Intercept)")`.

src/terms.jl
```diff
@@ -543,18 +543,48 @@ vectorize(x) = [x]
...
-    kron_insideout((args...) -> join(args, " & "), vectorize.(coefnames.(t.terms))...)
+    Symbol.(kron_insideout((args...) -> join(args, " & "), vectorize.(coefnames.(t.terms))...))
+StatsBase.coefnames(ts::TupleTerm) = _coefnames(ts.terms)
+_coefnames(ts::Tuple) = (coefnames(first(ts)), _coefnames(tail(ts))...)
+_coefnames(ts::Tuple{}) = ()
+
+"""
+    coef(term::AbstractTerm, s::Symbol)
+"""
+function StatsBase.coef(f::FormulaTerm, s::Symbol)
```
kleinschmidt: I don't see what the point of this is, or rather why it's called `coef` (which means something different in statsbase).  can you explain what your goal is here?  in any case, this needs tests if it's going to be included, and it seems like it would be better in a separate PR.

also, it doesn't seem like this works for categorical terms or for interactions, even if you specify the full coefficient name, since `coefnames` will be a vector of symbols.

and the RHS can have other things than just a `MatrixTerm` (e.g., MixedModels has a tuple of fixed and random effect matrices which are of different types)

kleinschmidt:
One nice consequence of this would be is that it would be easier to create named tuples from various kinds of model outputs (I just found myself converting coefnames into Symbols to do this).

nalimilan:
Let's discuss whether it's a good idea or not at https://github.com/JuliaStats/StatsModels.jl/issues/113.

Tokazama:
Thanks for taking a look at this. I'll address the issues if people agree it's a good idea. It was admittedly a messy first attempt to see how practical implementing it would be

Issue #164, opened 2019-11-10
Speed when creating formula


matthieugomez:
On my computer, it takes 0.3s to create a formula with functions in the REPL:
```
using StatsModels
 @time @formula(y ~ x+y)
#  0.000022 seconds (16 allocations: 912 bytes)
@time @formula(y ~ log(x) + log(y))
#  0.355020 seconds (1.13 M allocations: 57.160 MiB, 6.92% gc time)
```
This sounds a bit excessive. Is there any way to make this faster?

oxinabox:
I strongly recommend timing using BenchmarkTool's `@btime`.
Which will properly ignore compile time* 
and is resistant to noise:
On julia 1.3-rc4
i get:
```
julia> @btime @formula(y ~ x+y);
  452.624 ns (12 allocations: 752 bytes)

julia> @btime @formula(y ~ log(x) + log(y));
  10.219 μs (72 allocations: 4.88 KiB)
```

*(If compile time is the issue then that should be highlighted since solutions are very different)

matthieugomez:
I know that `@btime` returns a smaller time, but that's not the point.
My point is that it takes way too much time to create a formula in the REPL.

kleinschmidt:
Any ideas why global scope would be slower?

kleinschmidt:
It does seem to only happen in global/REPL scope: 

```julia
julia> g() = @formula(y ~ log(x) + log(y) + log(z))
g (generic function with 1 method)

julia> @time g()
  0.366315 seconds (1.05 M allocations: 51.787 MiB, 1.78% gc time)
FormulaTerm
Response:
  y(unknown)
Predictors:
  (x)->log(x)
  (y)->log(y)
  (z)->log(z)

julia> @time g()
  0.000323 seconds (98 allocations: 6.109 KiB)
FormulaTerm
Response:
  y(unknown)
Predictors:
  (x)->log(x)
  (y)->log(y)
  (z)->log(z)

julia> @time @formula(y ~ log(x) + log(y) + log(z))
  0.342751 seconds (958.20 k allocations: 47.005 MiB)
FormulaTerm
Response:
  y(unknown)
Predictors:
  (x)->log(x)
  (y)->log(y)
  (z)->log(z)

julia> @time @formula(y ~ log(x) + log(y) + log(z))
  0.354228 seconds (958.19 k allocations: 47.031 MiB, 2.29% gc time)
FormulaTerm
Response:
  y(unknown)
Predictors:
[...]

PR #172, opened 2020-02-18
improved hypothesis coding for contrasts

kleinschmidt:
This adds or improves a few features for dealing with hypothesis-coded
contrasts:

* a (mandatory) `labels` field (and specialized `termnames` method) which allows
  custom labels for the generated contrasts.
* widen types for contrasts/hypothesis matrices so any `AbstractMatrix` can be
  supplied
* widen types on `ContrastsMatrix` so that levels and termnames don't need to
  have the same type (to support e.g. string labels for symbol-valued variable)
* remove the `base` field since it's meaningless for this type, and use an
  (un-exported) accessor function `baselevel` instead.
* an experimental/un-exported `hypothesis_matrix` function that shows the
  hypothesis matrix/interpretation matrix for any contrasts.  this is probably
  the roughest part so feedback is especially welcome


kleinschmidt:
Actually, as is, this would be a breaking change (since `HypothesisCoding` is already merged without mandatory labels).  So would be better to make missing labels a warning...and an error in a future release.

src/contrasts.jl
```diff
@@ -466,49 +474,79 @@ expressed as a hypothesis matrix, but it is not obvious just from looking at the
 contrasts matrix.  For this reason `HypothesisCoding` is preferred for
 specifying custom contrast coding schemes over `ContrastsCoding`.
 
+Optional arguments `levels` and `labels` give the ordering of mapping between
```
nalimilan: "ordering of mapping" isn't very clear to me. Basically, these are the column and row names of the hypothesis matrix, right?

`levels` and `labels` reminds me of the corresponding arguments to `factor` in R, which I always found confusing. Maybe we can find something more explicit than `labels`? Maybe `contrastlabels`?

(BTW, it seems like `NamedArray` inputs could be allowed at some point.)

src/contrasts.jl
```diff
@@ -466,49 +474,79 @@ expressed as a hypothesis matrix, but it is not obvious just from looking at the
...
-    hypotheses::Matrix
-    contrasts::Matrix
-    base::Nothing
-    levels::Union{Vector,Nothing}
-
-    function HypothesisCoding(hypotheses, base, levels)
+mutable struct HypothesisCoding{T<:AbstractMatrix, S<:AbstractMatrix} <: AbstractContrasts
+    hypotheses::T
+    contrasts::S
+    levels::Union{AbstractVector,Nothing}
```
nalimilan: Why parameterize on `S` and `T` but not on the type of this field?

src/contrasts.jl
```diff
@@ -466,49 +474,79 @@ expressed as a hypothesis matrix, but it is not obvious just from looking at the
...
+    labels::AbstractVector
+
+    function HypothesisCoding(hypotheses::T, levels, labels) where {T}
+        labels == nothing &&
+            throw(ArgumentError("must specify contrast labels with " *
+                                "HypothesisCoding(; labels=) or " *
```
nalimilan:
```suggestion
                                "HypothesisCoding(hypotheses; labels=...) or " *
```

src/contrasts.jl
```diff
@@ -466,49 +474,79 @@ expressed as a hypothesis matrix, but it is not obvious just from looking at the
...
     end
 end
 
-HypothesisCoding(mat::Matrix; levels=nothing) =
-    HypothesisCoding(mat, nothing, levels)
+HypothesisCoding(mat::AbstractMatrix; levels=nothing, labels=nothing) =
+    HypothesisCoding(mat, levels, labels)
+
+"""
+    HypothesisCoding(hypotheses::Dict[; labels=, levels=])
```
nalimilan: I think this is a more standard convention:
```suggestion
    HypothesisCoding(hypotheses::Dict[; labels::AbstractVector, levels::AbstractVector])
```
Otherwise you could show the defaults `labels=collect(keys(hypotheses)), levels=nothing`, which isn't too bad.

src/contrasts.jl
```diff
@@ -466,49 +474,79 @@ expressed as a hypothesis matrix, but it is not obvious just from looking at the
...
+
+Specify hypotheses as `label=>hypothesis_vector` pairs.  If labels are specified
+via keyword argument, the hypothesis vectors will be concatenated in that order.
+"""
+function HypothesisCoding(
```
nalimilan: I don't think this indentation style is used elsewhere in the package?

src/contrasts.jl
```diff
@@ -466,49 +474,79 @@ expressed as a hypothesis matrix, but it is not obvious just from looking at the
...
 
-HypothesisCoding(mat::Matrix; levels=nothing) =
-    HypothesisCoding(mat, nothing, levels)
+HypothesisCoding(mat::AbstractMatrix; levels=nothing, labels=nothing) =
+    HypothesisCoding(mat, levels, labels)
+
+"""
+    HypothesisCoding(hypotheses::Dict[; labels=, levels=])
+
+Specify hypotheses as `label=>hypothesis_vector` pairs.  If labels are specified
```
nalimilan: Would it make sense to take a vector of pairs instead? That would avoid repeating the labels twice.

Also the role of `levels` should be mentioned.

src/contrasts.jl
```diff
@@ -466,49 +474,79 @@ expressed as a hypothesis matrix, but it is not obvious just from looking at the
...
+"""
+function HypothesisCoding(
+    hypotheses::Dict{<:Any,<:AbstractVector};
+    labels=collect(keys(hypotheses)),
+    levels=nothing
+)
+    !isempty(symdiff(keys(hypotheses), labels)) &&
+        throw(ArgumentError("labels mismatch between hypotheses " *
+                            "($(collect(keys(hypotheses)))) " *
+                            "and labels kw arg ($labels)"))
```
nalimilan:
```suggestion
                            "and labels keyword argument ($labels)"))
```

src/contrasts.jl
```diff
@@ -466,49 +474,79 @@ expressed as a hypothesis matrix, but it is not obvious just from looking at the
...
+via keyword argument, the hypothesis vectors will be concatenated in that order.
+"""
+function HypothesisCoding(
+    hypotheses::Dict{<:Any,<:AbstractVector};
+    labels=collect(keys(hypotheses)),
+    levels=nothing
+)
+    !isempty(symdiff(keys(hypotheses), labels)) &&
+        throw(ArgumentError("labels mismatch between hypotheses " *
+                            "($(collect(keys(hypotheses)))) " *
```
nalimilan: I think you need to use `join` to get a nice result:
```julia
julia> "$(collect(keys(Dict("a"=>1, "b"=>2))))"
"[\"b\", \"a\"]"
```
Same for `labels`.

Also, maybe only print the result of `symdiff`? That will directly show the problem.

src/contrasts.jl
```diff
@@ -466,49 +474,79 @@ expressed as a hypothesis matrix, but it is not obvious just from looking at the
...
+                            "and labels kw arg ($labels)"))
+    
+    mat = reduce(hcat, hypotheses[label] for label in labels)'
```
nalimilan: `reduce(hcat, ...)` is only optimized for vector inputs. So you need to choose between allocating a copy of the input, and using the slow `reduce` which makes N copies.
```suggestion
    mat = reduce(hcat, collect(hypotheses[label] for label in labels))'
```

src/contrasts.jl
```diff
@@ -466,49 +474,79 @@ expressed as a hypothesis matrix, but it is not obvious just from looking at the
...
+    HypothesisCoding(mat; labels=labels, levels=levels)
+end
+
 
 function contrasts_matrix(C::HypothesisCoding, baseind, n)
     check_contrasts_size(C.contrasts, n)
     C.contrasts
 end
 
+termnames(C::HypothesisCoding, levels::AbstractVector, baseind::Int) =
```
nalimilan: Shouldn't this be `labels` instead of `levels`?

src/contrasts.jl
```diff
@@ -466,49 +474,79 @@ expressed as a hypothesis matrix, but it is not obvious just from looking at the
...
+    something(C.labels, levels[1:length(levels) .!= baseind])
 
 """
-    StatsModels.ContrastsCoding(mat::Matrix[, base[, levels]])
+    StatsModels.ContrastsCoding(mat::AbstractMatrix[, base[, levels]])
```
nalimilan:
```suggestion
    StatsModels.ContrastsCoding(mat::AbstractMatrix[; levels::AbstractVector])
```

`levels` should also be mentioned in the docstring.

src/contrasts.jl
```diff
@@ -466,49 +474,79 @@ expressed as a hypothesis matrix, but it is not obvious just from looking at the
...
 """
-mutable struct ContrastsCoding <: AbstractContrasts
-    mat::Matrix
-    base::Any
+mutable struct ContrastsCoding{T<:AbstractMatrix} <: AbstractContrasts
+    mat::T
     levels::Union{Vector,Nothing}
 
-    function ContrastsCoding(mat, base, levels)
+    function ContrastsCoding(mat::T, levels) where {T}
```
nalimilan:
```suggestion
    function ContrastsCoding(mat::T, levels) where {T<:AbstractMatrix}
```
Also same remark about using a keyword argument here.

src/contrasts.jl
```diff
@@ -520,10 +558,29 @@ check_contrasts_size(mat::Matrix, n_lev::Int) =
...
+ContrastsCoding(mat::AbstractMatrix; levels=nothing) =
+    ContrastsCoding(mat, levels)
 
 function contrasts_matrix(C::ContrastsCoding, baseind, n)
     check_contrasts_size(C.mat, n)
     C.mat
 end
+
+## hypothesis matrix
+needs_intercept(mat) =
```
nalimilan:
```suggestion
needs_intercept(mat::AbstractMatrix) =
```

src/contrasts.jl
```diff
@@ -520,10 +558,29 @@ check_contrasts_size(mat::Matrix, n_lev::Int) =
...
+    (rank(mat) < size(mat, 1)) &&
+    !all(mapslices(sum, mat, dims=1) .< 10eps(eltype(mat)))
```
nalimilan: IIRC `mapslices` makes a copy of each slice, so better use `eachslice`.
```suggestion
    !all(sum(row) < 10eps(eltype(mat)) for row in eachrow(x))
```

src/contrasts.jl
kleinschmidt: well in an ideal world yes but this is the interface that `ContrastsMatrix` uses.  The arguments are just ignored in this method...

src/contrasts.jl
```diff
@@ -466,49 +474,79 @@ expressed as a hypothesis matrix, but it is not obvious just from looking at the
...
-    levels::Union{Vector,Nothing}
-
-    function HypothesisCoding(hypotheses, base, levels)
+mutable struct HypothesisCoding{T<:AbstractMatrix, S<:AbstractMatrix} <: AbstractContrasts
+    hypotheses::T
+    contrasts::S
+    levels::Union{AbstractVector,Nothing}
+    labels::AbstractVector
+
+    function HypothesisCoding(hypotheses::T, levels, labels) where {T}
```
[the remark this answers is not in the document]
kleinschmidt: there are kwarg methods further down, this is just to ensure that you can't construct a HypothesisCoding with mismatching contrasts and hypothesis matrices...

src/contrasts.jl
```diff
@@ -224,6 +231,8 @@ for contrastType in [:DummyCoding, :EffectsCoding, :HelmertCoding, :SeqDiffCodin
     end
 end
 
+baselevel(c::C) where {C<:AbstractContrasts} = hasfield(C, :base) ? c.base : nothing
```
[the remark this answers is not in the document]
kleinschmidt: Yeah it is hacky.  But you can always provide a more specific method than this for other types so the user-facing API wouldn't change at all.

src/contrasts.jl
kleinschmidt: ...and same thing about there being a separate kwargs constructor defined outside.

src/contrasts.jl
kleinschmidt: yeah NamedArray would be nice.  but is that a dependency we can take?

`contrastlabels` is a little verbose to my taste so maybe we should get feedback from others on whether `labels` vs. `levels` is clear

src/contrasts.jl
kleinschmidt: Great question.  It didn't seem like it would buy anything performance wise since it's going through the ContrastsMatrix anyway before it's used, and I was starting to get tired of so many type parameters...

src/contrasts.jl
kleinschmidt: I like the more verbose version better actually.

src/contrasts.jl
kleinschmidt: So that would be another method for `::Vector{Pair}`?

src/contrasts.jl
kleinschmidt: ah yeah I've just gotten used to it else where...I'll make it consistent with what we are doing elsewhere rather than changing the whole package to use BlueStyle.jl or whatever it's called

src/contrasts.jl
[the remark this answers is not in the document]
nalimilan: Yeah, but then if one day you wanted to rename this field, it could break external implementations. That's not very likely, but better avoid exposing internal fields.

src/contrasts.jl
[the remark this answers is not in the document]
nalimilan: But couldn't there just be one single (inner) constructor taking keyword arguments?

src/contrasts.jl
nalimilan: > yeah NamedArray would be nice. but is that a dependency we can take?

Probably not (not at this point at least), I just wanted to mention that.

src/contrasts.jl
nalimilan: OK, I just wanted to check there was a (good) reason.

src/contrasts.jl
nalimilan: Yeah, or it could even replace the `::Dict` method.

src/contrasts.jl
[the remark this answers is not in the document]
kleinschmidt: hm, maybe.  I hadn't thought about that.  I'm not sure we want to force people to use keyword arguments but honestly I don't have a strong opinion about it..

src/contrasts.jl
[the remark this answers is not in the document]
kleinschmidt: Ah I see now.  Yes better to have an accessor for the types that have the `base` field and a separate fallback...

src/contrasts.jl
kleinschmidt: seems like the dependencies are pretty light, but not nothing (recipesbase, combinatorics, requires).

src/contrasts.jl
kleinschmidt: isn't the dict the paradigmatic key-value map? it seems weird to require a vector of pairs over a dict.  but then again pairs do seem to crop up more and more (like in the dataframes stuff)

kleinschmidt:
Turns out that `eachcol` and `eachslice` aren't available on 1.0...so maybe we need to do `mapslices` even if its a bit slower.  Anyway I don't think it's in a hot path so performance isn't the worst.  It also just occurred to me that `sum(...dims=2)` should work so I've done that...

src/contrasts.jl
nalimilan: I think the main issue is that the stats ecosystem hasn't really standardized on a package for arrays with dimension names, due to the NamedArrays vs. AxisArrays divide.

src/contrasts.jl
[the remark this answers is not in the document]
nalimilan: In general AFAICT APIs choose either positional or keyword arguments, but don't support both.

src/contrasts.jl
nalimilan: Though we could imagine defining a set of common functions (e.g. in DataAPI) so that other packages can use names without depending on a particular array package.

src/contrasts.jl
nalimilan: Well, yeah, but since `Dict` doesn't preserve order it's often not what we need. `Vector{<:Pair}` is fine as long as you don't need to lookup keys among a large number of entries (which isn't the case here).

nalimilan:
> Turns out that `eachcol` and `eachslice` aren't available on 1.0...so maybe we need to do `mapslices` even if its a bit slower. Anyway I don't think it's in a hot path so performance isn't the worst. It also just occurred to me that `sum(...dims=2)` should work so I've done that...

I think they are in Compat. If you don't want to add that dependency, you can also simply do something like `(sum(view(x, :, i)) for i in 1:size(x, 2))`. Otherwise, `sum(..., dims=2)` is much better than `mapslices` (it will allocate a vector to hold the sums, which will be much smaller than allocating slices, and it will happen only once).

src/contrasts.jl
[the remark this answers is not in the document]
kleinschmidt: I see, and I agree that the kwargs are a better api.  But changing that would be a breaking change; there are a few contrasts-related breaking changes (keyword args, error on missing levels for hypothesis, type parameters) so maybe we should merge something that's non-breaking now and then make another breaking release.  Or just give up and make a breaking release next?  we're not on 1.0 yet so that's not SO disruptive I guess.

src/contrasts.jl
kleinschmidt: Ah I see.  I'd like to use something like a NamedArray as teh internal representation of the contrasts matrix, or at least for displaying.  I'd really like to be able to print the contrasts/hypothesis matrices with nice row/column labels but also really don't feel like re-inventing the wheel for that.

src/contrasts.jl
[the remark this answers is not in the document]
nalimilan: Ah, right, I guess that's fine then. Maybe just leave an issue with the 1.0 milestone so that we decide whether to deprecate positional arguments at that point.

src/contrasts.jl
nalimilan: Yeah it's really annoying that we haven't standardized on something yet.

src/contrasts.jl
```diff
@@ -500,22 +543,22 @@ HypothesisCoding(mat::AbstractMatrix; levels=nothing, labels=nothing) =
...
+                          levels=nothing)
     !isempty(symdiff(keys(hypotheses), labels)) &&
-        throw(ArgumentError("labels mismatch between hypotheses " *
-                            "($(collect(keys(hypotheses)))) " *
-                            "and labels kw arg ($labels)"))
+        throw(ArgumentError("mismatching labels between hypotheses and labels keyword argument: " *
+                            "$(join(symdiff(keys(hypotheses), labels), ", "))"))
     
-    mat = reduce(hcat, hypotheses[label] for label in labels)'
+    mat = reduce(hcat, collect(hypotheses[label] for label in labels))'
```
nalimilan:
```suggestion
    mat = reduce(hcat, [hypotheses[label] for label in labels])'
```

src/contrasts.jl
```diff
@@ -567,20 +614,94 @@ function contrasts_matrix(C::ContrastsCoding, baseind, n)
...
+    !all(isapprox.(sum(mat, dims=1), 0, rtol=1e-5))
 
-hypothesis_matrix(contrasts::AbstractContrasts, baseind, n; kwargs...) =
-    hypothesis_matrix(contrasts_matrix(contrasts, baseind, n); kwargs...)
+"""
+    hypothesis_matrix(cmat::AbstractMatrix; intercept=needs_intercept(cm), pretty=true)
+    hypothesis_matrix(contrasts::AbstractContrasts, n; baseind=1, kwargs...)
+    hypothesis_matrix(cmat::ContrastsMatrix; kwargs...)
+
+Compute the hypotehsis matrix for a contrasts matrix using the generalized
```
nalimilan:
```suggestion
Compute the hypothesis matrix for a contrasts matrix using the generalized
```

src/contrasts.jl
```diff
@@ -567,20 +614,94 @@ function contrasts_matrix(C::ContrastsCoding, baseind, n)
...
+pseudo-inverse (`LinearAlgebra.pinv`).  `intercept` determines whether a column
+of ones is included before taking the pseudoinverse, which is needed for
+contrasts where the columns are not orthogonal to the intercept (e.g., have
+non-zero mean).  If `pretty=true` (the default), the hypotheses are rounded to
+`Int`s if possible and `Rational`s if not, using a tolerance of `1e-10`.
+
+Note that this assumes a *balanced design* where there are the same number of
+observations in every cell.  This is only important for non-orthgonal contrasts
+(including contrasts where the contrasts are not orthogonal with the intercept).
```
nalimilan:
```suggestion
(including contrasts that are not orthogonal with the intercept).
```

src/contrasts.jl
```diff
@@ -567,20 +614,94 @@ function contrasts_matrix(C::ContrastsCoding, baseind, n)
...
 
+hypothesis_matrix(contrasts::AbstractContrasts, n; baseind=1, kwargs...) =
+    hypothesis_matrix(contrasts_matrix(contrasts, baseind, n); kwargs...)
+
+hypothesis_matrix(cmat::ContrastsMatrix; kwargs...) =
+    hypothesis_matrix(cmat.matrix; kwargs...)
+
+function pretty_mat(mat::AbstractMatrix; tol=10*eps(eltype(mat)))
+    fracs = rationalize.(mat, tol=tol)
+    if all(denominator.(fracs) .== 1)
```
nalimilan: So this is completely type-unstable. I guess you assume it's not a problem for performance?

Also this would avoid a few allocations:
```suggestion
    if all(x -> denominator(x) == 1, fracs)
```

src/contrasts.jl
```diff
@@ -567,20 +614,94 @@ function contrasts_matrix(C::ContrastsCoding, baseind, n)
...
+"""
+    hypothesis_matrix(cmat::AbstractMatrix; intercept=needs_intercept(cm), pretty=true)
+    hypothesis_matrix(contrasts::AbstractContrasts, n; baseind=1, kwargs...)
+    hypothesis_matrix(cmat::ContrastsMatrix; kwargs...)
+
+Compute the hypotehsis matrix for a contrasts matrix using the generalized
+pseudo-inverse (`LinearAlgebra.pinv`).  `intercept` determines whether a column
+of ones is included before taking the pseudoinverse, which is needed for
+contrasts where the columns are not orthogonal to the intercept (e.g., have
+non-zero mean).  If `pretty=true` (the default), the hypotheses are rounded to
```
nalimilan: Instead of `pretty`, this could be a `tolerance` argument defaulting to `0`?

src/contrasts.jl
kleinschmidt: oh dear :disappointed:

src/contrasts.jl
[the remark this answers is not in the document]
kleinschmidt: 1.0 seems a little ambitious at this point but I made a 0.7 :)

src/contrasts.jl
kleinschmidt: I've thought a bit more about it and I agree.  I'll make the change...

src/contrasts.jl
kleinschmidt: good idea

kleinschmidt: actually, I'd rather have a nonzero tolerance, to default to pretty-printing the matrix.  and setting it to 0 would opt out of rationalizing it.

src/contrasts.jl
kleinschmidt: yeah it's not meant to be used for performance-critical things, more as a helpful visualization/inspection tool.  I wonder if using pretty=true/false vs. tolerance=0/nonzero makes a difference there?

nalimilan: Good question. I suspect that the default will be optimized away, but that could be worth checking.

kleinschmidt:
Ah sorry I thought I had; I'm thinking of `hypothesis_matrix` as more of a way to inspect the hypotheses that would be generated given a particular contrasts matrix, rather than as an interface for creating an instance of `HypothesisCoding` (which requires more information like labels for the hypotheses...)

nalimilan:
As you like!

kleinschmidt:
actually, there's a problem with this PR which is that it changes the type parameters on ContrastsMatrix, which might be considered breaking (and could in practice lead to type instability for accessing the termnames, which might not be a HUGE deal for performance but I'm not sure...).  Also it introduces type parameters on hypothesis and contrasts coding structs as well.  So maybe this just has to be a breaking change and we can be more aggressive about changing the API as well...

nalimilan:
I think whether adding a type parameter is breaking should be appreciated based on actual/typical usage. Do you think any users would really be affected in terms of performance?

kleinschmidt:
I ran a quick benchmarking test of my intuitions that this could make a performance difference (because the contrasts are a type parameter on the contrastsmatrix, and the categorical term, but the categorical term's parameter is the contrasts and the levels type but not the added termnames parameter...).  I see a reliable but minimal difference for `modelcols` on a large table, and a larger difference when looking at a single row at a time (10-100% depending on the test). I suspect that most users are this point are not doing row-wise stuff, so it's likely okay.  And we need to do a performance overhaul anyway (cf. #164), so might be okay to just stick with this for now.  The alternative would be to change the type parameters on the categorical term in this PR as well.

kleinschmidt:
(but I'm leaning towards a "type-parameter-pocolypse" in the near future anyway so maybe best to just leave it for now)

src/contrasts.jl
```diff
@@ -221,9 +241,13 @@ for contrastType in [:DummyCoding, :EffectsCoding, :HelmertCoding, :SeqDiffCodin
         end
         ## constructor with optional keyword arguments, defaulting to nothing
         $contrastType(; base=nothing, levels=nothing) = $contrastType(base, levels)
+        baselevel(c::$contrastType) = c.base
     end
 end
 
+# fallback method for other types that might not have base field
+baselevel(c::C) where {C<:AbstractContrasts} = nothing
```
nalimilan:
```suggestion
baselevel(c::AbstractContrasts) = nothing
```

src/contrasts.jl
```diff
@@ -253,18 +277,23 @@ mutable struct FullDummyCoding <: AbstractContrasts
...
     DummyCoding([base[, levels]])
+    DummyCoding(; base=nothing, levels=nothing)
 
 Dummy coding generates one indicator column (1 or 0) for each non-base level.
 
+If `levels` are omitted or `nothing`, they are determined from the data using
+by calling the `levels` function on the data when constructing `ContrastsMatrix`.
+If `base` is omitted or
+`nothing`, the first level is used as the base.
+
```
nalimilan:
```suggestion
If `levels` are omitted or `nothing`, they are determined from the data
by calling the `levels` function on the data when constructing `ContrastsMatrix`.
If `base` is omitted or `nothing`, the first level is used as the base.

```

src/contrasts.jl
```diff
@@ -299,6 +329,10 @@ column is generated with 1 where `variable .== x` and -1 where `variable .== bas
 `EffectsCoding` is like `DummyCoding`, but using -1 for the base level instead
 of 0.
 
+If `levels` are omitted or `nothing`, they are determined from the data using
+`levels()` when constructing `Contrastsmatrix`.  If `base` is omitted or
+`nothing`, the first level is used as the base.
```
nalimilan:
```suggestion
If `levels` are omitted or `nothing`, they are determined from the data
by calling the `levels` function when constructing `ContrastsMatrix`. 
If `base` is omitted or `nothing`, the first level is used as the base.
```

src/contrasts.jl
```diff
@@ -332,10 +366,15 @@ end
...
     HelmertCoding([base[, levels]])
+    HelmertCoding(; base=nothing, levels=nothing)
 
 Helmert coding codes each level as the difference from the average of the lower
 levels.
 
+If `levels` are omitted or `nothing`, they are determined from the data using
+`levels()` when constructing `Contrastsmatrix`.  If `base` is omitted or
+`nothing`, the first level is used as the base.
+
```
nalimilan:
```suggestion
If `levels` are omitted or `nothing`, they are determined from the data
by calling the `levels` function when constructing `Contrastsmatrix`. 
If `base` is omitted or `nothing`, the first level is used as the base.
```

src/contrasts.jl
```diff
@@ -375,6 +414,11 @@ compares each level to the level below it (starting with the second level).
 Specifically, the ``n``th predictor tests the hypothesis that the difference
 between levels ``n`` and ``n+1`` is zero.
 
+Differences are computed in order of `levels`.  If `levels` are omitted or
+`nothing`, they are determined from the data using `levels()` when constructing
+`Contrastsmatrix`.  If `base` is omitted or `nothing`, the first level is used
+as the base.
```
nalimilan:
```suggestion
`nothing`, they are determined from the data by calling the `levels` function
when constructing `ContrastsMatrix`.
If `base` is omitted or `nothing`, the first level is used as the base.
```

src/contrasts.jl
```diff
@@ -466,49 +513,93 @@ expressed as a hypothesis matrix, but it is not obvious just from looking at the
...
+of Memory and Language, 110_, 104038. https://doi.org/10.1016/j.jml.2019.104038
+
+"""
+mutable struct HypothesisCoding{T<:AbstractMatrix, S<:AbstractMatrix} <: AbstractContrasts
+    hypotheses::T
+    contrasts::S
+    levels::Union{AbstractVector,Nothing}
+    labels::Union{AbstractVector,Nothing}
+
+    function HypothesisCoding(hypotheses::T, levels, labels) where {T}
```
nalimilan:
```suggestion
    function HypothesisCoding(hypotheses::T, levels::Union{AbstractVector,Nothing},
                              labels::Union{AbstractVector,Nothing}) where {T}
```

src/contrasts.jl
```diff
@@ -466,49 +513,93 @@ expressed as a hypothesis matrix, but it is not obvious just from looking at the
...
+
+Specify hypotheses as `label=>hypothesis_vector` pairs.  If labels are specified
+via keyword argument, the hypothesis vectors will be concatenated in that order.
+The `levels` argument provides the names of the levels for each entry in the
+hypothesis vectors.  If omitted, the `levels` function will be called on the data when constructing a 
+`ContrastsMatrix`
+"""
+function HypothesisCoding(hypotheses::Dict{<:Any,<:AbstractVector};
+                          labels=collect(keys(hypotheses)),
+                          levels=nothing)
```
nalimilan:
```suggestion
                          labels::Union{AbstractVector,Nothing}=collect(keys(hypotheses)),
                          levels::Union{AbstractVector,Nothing}=nothing)
```

src/contrasts.jl
```diff
@@ -466,49 +513,93 @@ expressed as a hypothesis matrix, but it is not obvious just from looking at the
...
+        S = typeof(contrasts)
         check_contrasts_size(contrasts, levels)
-        new(hypotheses, contrasts, base, levels)
+        new{T,S}(hypotheses, contrasts, levels, labels)
     end
 end
 
-HypothesisCoding(mat::Matrix; levels=nothing) =
-    HypothesisCoding(mat, nothing, levels)
+HypothesisCoding(mat::AbstractMatrix; levels=nothing, labels=nothing) =
```
nalimilan:
```suggestion
HypothesisCoding(mat::AbstractMatrix;
                 levels::Union{AbstractVector,Nothing}=nothing,
                 labels::Union{AbstractVector,Nothing}=nothing) =
```

src/contrasts.jl
```diff
@@ -466,49 +513,93 @@ expressed as a hypothesis matrix, but it is not obvious just from looking at the
...
-    mat::Matrix
-    base::Any
+mutable struct ContrastsCoding{T<:AbstractMatrix} <: AbstractContrasts
+    mat::T
     levels::Union{Vector,Nothing}
 
-    function ContrastsCoding(mat, base, levels)
-        Base.depwarn("`ContrastsCoding(contrasts)` is deprecated and will not be exported" *
-                     " in the future, use `HypothesisCoding(pinv(contrasts))` instead.",
+    function ContrastsCoding(mat::T, levels) where {T<:AbstractMatrix}
```
nalimilan:
```suggestion
    function ContrastsCoding(mat::T, levels::Union{AbstractVector,Nothing}) where {T<:AbstractMatrix}
```

src/contrasts.jl
```diff
@@ -520,10 +611,104 @@ check_contrasts_size(mat::Matrix, n_lev::Int) =
                         "Expected $((n_lev, n_lev-1)), got $(size(mat))"))
 
 ## constructor with optional keyword arguments, defaulting to nothing
-ContrastsCoding(mat::Matrix; base=nothing, levels=nothing) =
-    ContrastsCoding(mat, base, levels)
+ContrastsCoding(mat::AbstractMatrix; levels=nothing) =
```
nalimilan:
```suggestion
ContrastsCoding(mat::AbstractMatrix; levels::Union{AbstractVector,Nothing}=nothing) =
```

src/contrasts.jl
```diff
@@ -520,10 +611,104 @@ check_contrasts_size(mat::Matrix, n_lev::Int) =
...
+    iszero(tolerance) ? hypotheses : pretty_mat(hypotheses, tol=tolerance)
+end
+
+hypothesis_matrix(contrasts::AbstractContrasts, n; baseind=1, kwargs...) =
+    hypothesis_matrix(contrasts_matrix(contrasts, baseind, n); kwargs...)
+
+hypothesis_matrix(cmat::ContrastsMatrix; kwargs...) =
+    hypothesis_matrix(cmat.matrix; kwargs...)
+
+function pretty_mat(mat::AbstractMatrix; tol=10*eps(eltype(mat)))
```
nalimilan:
```suggestion
function pretty_mat(mat::AbstractMatrix; tol::Real=10*eps(eltype(mat)))
```

src/contrasts.jl
```diff
@@ -520,10 +611,104 @@ check_contrasts_size(mat::Matrix, n_lev::Int) =
...
+julia> StatsModels.hypothesis_matrix(StatsModels.ContrastsMatrix(DummyCoding(), ["a", "b", "c", "d"]))
+4×4 Array{Int64,2}:
+ 1  -1  -1  -1
+ 0   1   0   0
+ 0   0   1   0
+ 0   0   0   1
+
+```
+"""
+function hypothesis_matrix(cm::AbstractMatrix; intercept=needs_intercept(cm), tolerance=1e-5)
```
nalimilan:
```suggestion
function hypothesis_matrix(cm::AbstractMatrix;
                           intercept::Bool=needs_intercept(cm), tolerance::Real=1e-5)
```

src/contrasts.jl
```diff
@@ -520,10 +611,104 @@ check_contrasts_size(mat::Matrix, n_lev::Int) =
...
+    if intercept
+        cm = hcat(ones(size(cm, 1)), cm)
+    end
+    hypotheses = pinv(cm)
+    iszero(tolerance) ? hypotheses : pretty_mat(hypotheses, tol=tolerance)
+end
+
+hypothesis_matrix(contrasts::AbstractContrasts, n; baseind=1, kwargs...) =
```
nalimilan:
```suggestion
hypothesis_matrix(contrasts::AbstractContrasts, n::Integer; baseind::Integer=1, kwargs...) =
```

src/contrasts.jl
```diff
@@ -466,49 +513,93 @@ expressed as a hypothesis matrix, but it is not obvious just from looking at the
...
+    HypothesisCoding(mat, levels, labels)
+
+"""
+    HypothesisCoding(hypotheses::Dict[; labels=collect(keys(hypotheses)), levels=nothing])
+
+Specify hypotheses as `label=>hypothesis_vector` pairs.  If labels are specified
+via keyword argument, the hypothesis vectors will be concatenated in that order.
+The `levels` argument provides the names of the levels for each entry in the
+hypothesis vectors.  If omitted, the `levels` function will be called on the data when constructing a 
+`ContrastsMatrix`
```
nalimilan:
```suggestion
`ContrastsMatrix`.
```

Issue #173, opened 2020-02-20
Override == and isequal to compare schemas


mrrobot-2000:
There is no built-in way in the package to check if two Schemas are equal.

PR #174, opened 2020-02-22
Overload == and isequal for checking equality of Terms and Schemas

mrrobot-2000:
#173


src/terms.jl
```diff
@@ -378,7 +418,7 @@ end
 
 Base.show(io::IO, t::InterceptTerm{H}) where {H} = print(io, H ? "1" : "0")
 
-Base.show(io::IO, t::ContinuousTerm) = print(io, t.sym)
+Base.show(io::IO, t::ContinuousTerm) = print(io, t.sym,"\n" , t.mean,"\n" , t.var, "\n" ,t.min,"\n" , t.max)
```
kleinschmidt: this is irrelevant for checking equality and breaks existing behavior.  It also causes the tests to fail.  As a general suggestion, make sure that the test suite runs (by doing `Pkg.test()` or running `test StatsModels` in pkg mode) before you submit a PR.

test/schema.jl
```diff
@@ -1,10 +1,17 @@
...
+    b = rand(9)
+
+    df = (y = y, a = 1:9, b = b, c = repeat(["d", "e", "f"], 3))
     f = apply_schema(f, schema(f, df))
     @test f == apply_schema(f, schema(f, df))
 
+    df2 = (y = y, a = 1:9, b = b, c = repeat(["d", "e", "f"], 3))
+
+    @test schema(df) == schema(df2)
+    @test isequal(schema(df), schema(df2))
```
kleinschmidt: this is a good start but you should also test as many of the "edge cases" you can think of as possible, since this will serve to make sure we don't break things in the future.  In particular, I'd suggest adding tests for 

* schema/terms with different named variables are not equal
* schema/terms with different invariants (mean/var/min/max, contrasts) are not equal
* schema are generated based on different tables with the same invariants (e.g., df1 with each column repeated twice) are equal

And whatever else you can dream up :)

mrrobot-2000:
Hi, Thanks for reviewing. I'll do the changes and add a summary of expected behaviour.

test/schema.jl
mrrobot-2000: I understood the first two points. Can you explain the last one in some detail ?

kleinschmidt: the schema shouldn't depend on the exact data that generated it, as long as the unique levels (for categorical) or mean/std/min/max (for continuous) are the same.  you check it right now by creating what is essentially a deep copy of the table, but it would be good to also check using a table that's not exactly the same size but has the same invariants, e.g. something like

`df2 = (y = [df1.y; df1.y], a = [1:9; 1:9], b = [b; b], c = [df.c; df.c])`

src/schema.jl
```diff
@@ -53,6 +53,29 @@ Base.merge!(a::Schema, b::Schema) = (merge!(a.schema, b.schema); a)
 Base.keys(schema::Schema) = keys(schema.schema)
 Base.haskey(schema::Schema, key) = haskey(schema.schema, key)
 
+function ==(first::Schema, second::Schema)
+    first === second && return true
+    first.schema === second.schema && return true
+    length(first.schema) != length(second.schema) && return false
+    for key in keys(first)
+        !haskey(second, key) && return false
+        get(second, key, nothing) != get(first, key, nothing) && return false
```
kleinschmidt: I don't think `get` is necessary here since you've already checked that the key is present on both sides.
```suggestion
        second[key] != first[key] && return false
```

src/schema.jl
```diff
@@ -53,6 +53,29 @@ Base.merge!(a::Schema, b::Schema) = (merge!(a.schema, b.schema); a)
...
+end
+
+function Base.isequal(first::Schema, second::Schema)
+    first === second && return true
+    first.schema === second.schema && return true
+    length(first.schema) != length(second.schema) && return false
+    for key in keys(first)
+        !haskey(second, key) && return false
+        !isequal(get(second, key, nothing), get(first, key, nothing)) &&
+        return false
```
kleinschmidt: Same here

```suggestion
        !isequal(second[key], first[key]) && return false
```

src/terms.jl
```diff
@@ -250,6 +286,13 @@ end
 MatrixTerm(t::AbstractTerm) = MatrixTerm((t, ))
 width(t::MatrixTerm) = sum(width(tt) for tt in t.terms)
 
+# ==(first::MatrixTerm, second::MatrixTerm) =
+#    collect_matrix_terms(first.terms) == collect_matrix_terms(second.terms)
+# isequal(first::MatrixTerm, second::MatrixTerm) =
+#    isequal(collect_matrix_terms(first.terms), collect_matrix_terms(second.terms))
```
kleinschmidt: explain why these are still here or delete

test/schema.jl
```diff
@@ -1,10 +1,39 @@
 @testset "schemas" begin
-
+    import Base
```
kleinschmidt: is this needed?

test/schema.jl
```diff
@@ -1,10 +1,39 @@
...
+    @test isequal(schema(df), schema(df2))
+    @test !isequal(schema(df), schema(df3))
+    @test !isequal(schema(df), schema(df4))
+    @test !isequal(schema(df), schema(df5))
+    @test !isequal(sch, sch2)
+
+    # @test schema(df) == schema(df3)
+    # @test isequal(schema(df), schema(df3))
+    #@test schema(df) != schema(df4)
+    #@test isequal(schema(df), schema(df4))
```
kleinschmidt: remove

test/schema.jl
```diff
@@ -1,10 +1,39 @@
...
+    b = rand(9)
+
+    df = (y = y, a = 1:9, b = b, c = repeat(["d", "e", "f"], 3))
     f = apply_schema(f, schema(f, df))
     @test f == apply_schema(f, schema(f, df))
 
+    df2 = (y = y, a = 1:9, b = b, c = [df.c; df.c])
+    df3 = (y = y, a = 1:9, b = b, c = repeat(["a", "b", "c"], 3))
+    df4 = (y = [df.y; df.y], a = [1:9; 1:9], b = [b; b], c = [df.c; df.c])
+    df5 = (z = y, a = 1:9, b = b, c = repeat(["d", "e", "f"], 3))
```
kleinschmidt: A few more cases to test: same values but different keys (both more/fewer keys in first/second, and equal numbers but different keys).

src/terms.jl
```diff
@@ -211,6 +233,14 @@ struct ContinuousTerm{T} <: AbstractTerm
 end
 width(::ContinuousTerm) = 1
 
+==(first::ContinuousTerm, second::ContinuousTerm) = first.sym == second.sym &&
```
kleinschmidt: consistent style:

```suggestion
==(first::ContinuousTerm, second::ContinuousTerm) = 
    first.sym == second.sym &&
```

src/terms.jl
```diff
@@ -187,6 +205,10 @@ via the [`implicit_intercept`](@ref) trait).
 struct InterceptTerm{HasIntercept} <: AbstractTerm end
 width(::InterceptTerm{H}) where {H} = H ? 1 : 0
 
+==(first::InterceptTerm, second::InterceptTerm) = width(first) == width(second)
```
kleinschmidt: consistent style
```suggestion
==(first::InterceptTerm, second::InterceptTerm) = 
    width(first) == width(second)
```

src/terms.jl
```diff
@@ -127,6 +135,12 @@ FunctionTerm(forig::Fo, fanon::Fa, names::NTuple{N,Symbol},
     FunctionTerm{Fo, Fa, names}(forig, fanon, exorig, args_parsed)
 width(::FunctionTerm) = 1
 
+==(first::FunctionTerm, second::FunctionTerm) = first.forig == second.forig &&
```
kleinschmidt:
```suggestion
==(first::FunctionTerm, second::FunctionTerm) =
    first.forig == second.forig &&
```

src/terms.jl
```diff
@@ -54,6 +57,11 @@ struct FormulaTerm{L,R} <: AbstractTerm
     rhs::R
 end
 
+==(first::FormulaTerm, second::FormulaTerm) =
+    first.lhs == second.lhs && first.rhs == second.rhs
+isequal(first::FormulaTerm, second::FormulaTerm) = isequal(first.lhs, second.lhs) &&
```
kleinschmidt:
```suggestion
isequal(first::FormulaTerm, second::FormulaTerm) =
    isequal(first.lhs, second.lhs) &&
```

mrrobot-2000:
I have fixed all things you asked except adding tests for Matrix terms. I do not know how to create a matrix term or how to include one in a formula. Can you please give an example of one such term ?

kleinschmidt:
They come from doing `apply_schema`, which creates them on the RHS of the formula.

kleinschmidt:
New tests still not passing.

mrrobot-2000:
Hi! I have fixed the broken tests. The latest build is now passing.

test/schema.jl
```diff
@@ -1,10 +1,47 @@
...
+    @test schema(df) == schema(df2)
+    @test apply_schema(f, schema(df)) == apply_schema(f, schema(df2))
+    @test schema(df) != schema(df3)
+    @test schema(df) != schema(df4)
+    @test schema(df) != schema(df5)
+    @test schema(df) != schema(df6)
+    @test schema(df) != schema(df7)
+    @test schema(df) != schema(df8)
+    @test schema(df8) != schema(df)
+    @test apply_schema(f, schema(df)) == apply_schema(f, schema(df5))
```
kleinschmidt: First I was confused because `schema(df) != schema(df5)`, so the result of applying them to the same formula shouldn't be equal. But then I realized that `f` has already had a schema applied to it above, so this isn't testing what I thought it was testing.  Which makes me wonder, what is the purpose of this test?  `apply_schema(@formula(y ~ 1 + a + log(b) + c + b&c), schema(df5))` would produce an error because `y` isn't found in `df5`.  But since you've already applied a schema with `y` in it above, you're actually just re-applying the schema terms that ARE found. I'm not sure that's a particularly useful test, unless you were specifically trying to test this edge case... so can you explain what the goal of this test is?

PR #176, opened 2020-02-22
programmatic construction for lead/lag

kleinschmidt:
Came up on slack: this adds methods for `lead` and `lag` that take a term and an optional number of steps and creates the appropriate `LeadLagTerm` object.  So you can do things like

```julia
f = term(:y) ~ lead(term(:x), 3)
```

which is equivalent to 

```julia
f2 = @formula(y ~ lead(x, 3))
```


kleinschmidt:
@oxinabox I'd like your feedback but can't request a review...

PR #179, opened 2020-05-15
Fix tests with DataFrames 0.21

nalimilan:
Keep support for 0.20 since it's very easy.
Add compatibility bounds to ensure this problem doesn't happen again.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -12,7 +12,9 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 
 [compat]
+CategoricalArrays = "0.7"
 DataAPI = "1.1"
+DataFrames = "0.20, 0.21"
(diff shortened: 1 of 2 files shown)


Project.toml
```diff
@@ -12,7 +12,9 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 
 [compat]
+CategoricalArrays = "0.7"
 DataAPI = "1.1"
+DataFrames = "0.20, 0.21"
```
nalimilan, line 17: I've checked that 0.19 doesn't work since it's incompatible with the allowed Tables versions.

kleinschmidt:
I didn't realize it was possible/kosher to include compat entries for packages that weren't dependencies, but otherwise lgtm.

nalimilan:
I didn't know that either before I tested it (I was afraid it would restrict DataFrames versions even when not testing, but that's not the case).

PR #181, opened 2020-06-26
use DataAPI.levels instead of levels field on contrasts, terms as hints

kleinschmidt:
EDIT: I've reverted the Grouping contrasts in favor of putting them in MixedModels.jl for now (see JuliaStats/MixedModels.jl#339).  I've renamed the PR to better reflect the scope of the changes now.

--------------------------

This introduces a new type of "contrasts" which is just a placeholder which
blocks the generation of the contrasts matrix.  This is useful in situations
where you have a variable with a large number of levels which will be used as a
grouping in some 3rd-party defined term.  The main use case I have in mind is
the random effects terms in
[MixedModels.jl](https://github.com/JuliaStats/MixedModels.jl), where the
grouping may have thousands of levels or more.  Because the contrasts matrix is
dense you can easily get an out-of-memory error.

With this PR, you can specify `Dict(:my_grouping_var => Grouping())` in the
hints when constructing the schema, and the contrasts matrix won't be
constructed but the levels will be extracted and stored, and the "inverse index"
(the level-to-index `Dict`) will be computed and stored as well.

This is a bit of a hack and I'm not 100% sure this needs to live in
StatsModels.jl but I'm throwing it up here to see what y'all think.  Ideally I'd
like to have a mechanism where a package can take control _before_ the schema is
generated, which would remove the need for abusing the hints like this.

There are also a few minor changes here as well that make it a bit easier to
take further control of the schema generation process (which helps in this
case): there's an added method so that when the `hint` for a term is already an
`AbstractTerm`, that term is just inserted directly into the schema.  And I
replaced the (implicit) requirement that all `AbstractContrasts` have a `levels`
field, with a method for `DataAPI.levels` instead (along the same lines as
replacing the mandatory `base` field with a `baselevel` accessor function).
This is technically breaking but I would be surprised if anyone is relying on
the current behavior (famous last words...)

### DIFF
--- a/src/contrasts.jl
+++ b/src/contrasts.jl
@@ -727,4 +727,30 @@ function pretty_mat(mat::AbstractMatrix; tol::Real=10*eps(eltype(mat)))
         return fracs
     end
 end
-    
+
+"""
+    struct Grouping <: StatsModels.AbstractContrasts end
+
+A placeholder type to indicate that a categorical variable is only used for
+grouping and not for contrasts.  When creating a [`CategoricalTerm`](@ref), this
+skips constructing the contrasts matrix which makes it robust to large numbers
+of levels, while still holding onto the vector of levels and constructing the
+level-to-index mapping (`invindex` field of the [`ContrastsMatrix`](@ref).).
+
+Note that calling `modelcols` on a `CategoricalTerm{Grouping}` is an error.
+This is intended only as a placeholder for other packages to build on.
+
+# Examples
+
+```julia
+julia> schema((; grp = string.(1:100_000)))
+# out-of-memory error
+
+julia> schema((; grp = string.(1:100_000)), Dict(:grp => Grouping()))
+"""
+struct Grouping <: StatsModels.AbstractContrasts
+end
+
+# return an empty matrix
+contrasts_matrix(::Grouping, baseind, n) = zeros(0,0)
+termnames(::Grouping, levels::AbstractVector, baseind::Integer) = levels
--- a/src/contrasts.jl
+++ b/src/contrasts.jl
@@ -727,30 +727,3 @@ function pretty_mat(mat::AbstractMatrix; tol::Real=10*eps(eltype(mat)))
         return fracs
     end
 end
-
-"""
-    Grouping <: AbstractContrasts
-
-A placeholder type to indicate that a categorical variable is only used for
-grouping and not for contrasts.  When creating a [`CategoricalTerm`](@ref), this
-skips constructing the contrasts matrix which makes it robust to large numbers
-of levels, while still holding onto the vector of levels and constructing the
-level-to-index mapping (`invindex` field of the [`ContrastsMatrix`](@ref).).
-
-Note that calling `modelcols` on a `CategoricalTerm{Grouping}` is an error.
-This is intended only as a placeholder for other packages to build on.
-
-# Examples
-
-```julia
-julia> schema((; grp = string.(1:100_000)))
-# out-of-memory error
-
-julia> schema((; grp = string.(1:100_000)), Dict(:grp => Grouping()))
-"""
-struct Grouping <: AbstractContrasts
-end
-
-# return an empty matrix
-contrasts_matrix(::Grouping, baseind, n) = zeros(0,0)
-termnames(::Grouping, levels::AbstractVector, baseind::Integer) = levels
--- a/src/contrasts.jl
+++ b/src/contrasts.jl
@@ -159,12 +159,12 @@ function ContrastsMatrix(contrasts::C, levels::AbstractVector{T}) where {C<:Abst
     # if levels are defined on contrasts, use those, validating that they line up.
     # what does that mean? either:
     #
-    # 1. contrasts.levels == levels (best case)
+    # 1. levels(contrasts) == levels (best case)
     # 2. data levels missing from contrast: would generate empty/undefined rows.
     #    better to filter data frame first
     # 3. contrast levels missing from data: would have empty columns, generate a
     #    rank-deficient model matrix.
-    c_levels = something(contrasts.levels, levels)
+    c_levels = something(levels(contrasts), levels)
     if eltype(c_levels) != eltype(levels)
         throw(ArgumentError("mismatching levels types: got $(eltype(levels)), expected " *
                             "$(eltype(c_levels)) based on contrasts levels."))
@@ -242,11 +242,13 @@ for contrastType in [:DummyCoding, :EffectsCoding, :HelmertCoding, :SeqDiffCodin
         ## constructor with optional keyword arguments, defaulting to nothing
         $contrastType(; base=nothing, levels::Union{AbstractVector,Nothing}=nothing) = $contrastType(base, levels)
         baselevel(c::$contrastType) = c.base
+        DataAPI.levels(c::$contrastType) = c.levels
     end
 end
 
-# fallback method for other types that might not have base field
+# fallback method for other types that might not have base or level fields
 baselevel(c::AbstractContrasts) = nothing
+DataAPI.levels(c::AbstractContrasts) = nothing
 
 """
     FullDummyCoding()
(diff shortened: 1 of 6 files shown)


palday:
I'm tempted to suggest that we add this into MixedModels for the time being (so that we have this available before the big 3.0 release) and if this gets integrated anywhere, then we just change our internal representation to issue a deprecation warning and call the canonical home.

dmbates:
Agree with @palday that we may want to copy this to MixedModels for the 3.0.0 release and reconcile later if needed.

src/contrasts.jl
```diff
@@ -159,12 +159,12 @@ function ContrastsMatrix(contrasts::C, levels::AbstractVector{T}) where {C<:Abst
...
     # what does that mean? either:
     #
-    # 1. contrasts.levels == levels (best case)
+    # 1. DataAPI.levels(contrasts) == levels (best case)
     # 2. data levels missing from contrast: would generate empty/undefined rows.
     #    better to filter data frame first
     # 3. contrast levels missing from data: would have empty columns, generate a
     #    rank-deficient model matrix.
-    c_levels = something(contrasts.levels, levels)
+    c_levels = something(DataAPI.levels(contrasts), levels)
```
nalimilan, line 167:
```suggestion
    c_levels = something(levels(contrasts), levels)
```

src/contrasts.jl
```diff
@@ -159,12 +159,12 @@ function ContrastsMatrix(contrasts::C, levels::AbstractVector{T}) where {C<:Abst
     # if levels are defined on contrasts, use those, validating that they line up.
     # what does that mean? either:
     #
-    # 1. contrasts.levels == levels (best case)
+    # 1. DataAPI.levels(contrasts) == levels (best case)
```
nalimilan, line 162:
```suggestion
    # 1. levels(contrasts) == levels (best case)
```

src/contrasts.jl
```diff
@@ -242,11 +242,13 @@ for contrastType in [:DummyCoding, :EffectsCoding, :HelmertCoding, :SeqDiffCodin
...
         $contrastType(; base=nothing, levels::Union{AbstractVector,Nothing}=nothing) = $contrastType(base, levels)
         baselevel(c::$contrastType) = c.base
+        DataAPI.levels(c::$contrastType) = c.levels
     end
 end
 
-# fallback method for other types that might not have base field
+# fallback method for other types that might not have base or level fields
 baselevel(c::AbstractContrasts) = nothing
+DataAPI.levels(c::AbstractContrasts) = nothing
```
nalimilan, line 251: Is this needed or here just in case? Not a big deal, but it doesn't really follow the DataAPI interface.

src/contrasts.jl
```diff
@@ -721,4 +727,30 @@ function pretty_mat(mat::AbstractMatrix; tol::Real=10*eps(eltype(mat)))
         return fracs
     end
 end
-    
+
+"""
+    struct Grouping <: StatsModels.AbstractContrasts end
```
nalimilan, line 732:
```suggestion
    Grouping <: AbstractContrasts
```

src/contrasts.jl
```diff
@@ -721,4 +727,30 @@ function pretty_mat(mat::AbstractMatrix; tol::Real=10*eps(eltype(mat)))
...
+
+# Examples
+
+```julia
+julia> schema((; grp = string.(1:100_000)))
+# out-of-memory error
+
+julia> schema((; grp = string.(1:100_000)), Dict(:grp => Grouping()))
+"""
+struct Grouping <: StatsModels.AbstractContrasts
```
nalimilan, line 751:
```suggestion
struct Grouping <: AbstractContrasts
```

src/contrasts.jl
```diff
@@ -721,4 +727,30 @@ function pretty_mat(mat::AbstractMatrix; tol::Real=10*eps(eltype(mat)))
...
+end
+
+# return an empty matrix
+contrasts_matrix(::Grouping, baseind, n) = zeros(0,0)
```
nalimilan, line 755: Does this have to be defined at all? And if so, better return `nothing`?

src/contrasts.jl
```diff
@@ -721,4 +727,30 @@ function pretty_mat(mat::AbstractMatrix; tol::Real=10*eps(eltype(mat)))
...
+termnames(::Grouping, levels::AbstractVector, baseind::Integer) = levels
```
nalimilan, line 756: Maybe also define `levels(::Grouping)`, by storing levels in the the `Grouping` struct? Couldn't that be useful for packages?

src/contrasts.jl
kleinschmidt: This is actually necessary because there's a argument named `levels`

src/contrasts.jl
kleinschmidt: same as above (although here it's just for clarifying the logic so doesn't need to run)

src/contrasts.jl
kleinschmidt: Well, some kind of "I don't have specified levels" is needed as a fallback, but it could be something else like an empty tuple.

src/contrasts.jl
kleinschmidt: It needs to be an actual matrix since the `ContrastsMatrix` struct is restricted to hold a matrix.

kleinschmidt:
I think at this point it makes more sense to put the `Grouping` struct into MixedModels so it doesn't hold up using it there.  It's not sufficiently specialized for actual grouping uses here so unless anyone has any objections I'll just revert that commit.  The other changes in this branch ARE useful and important though so I'd like to keep them.  

@Tokazama you <3'ed the original post, do you have a use for this as it currently is?  And maybe @Nosferican or @matthieugomez might have some interest?  Otherwise I'll revert.

Tokazama:
I don't plan on doing something with this in the immediate future. I like the idea of letting some of this mature in MixedModels and if it  later is needed can be back ported here.

matthieugomez:
I don't think I'd use something like this. These high dimensional categorical variables are so frequent for FixedEffectModels that it's just simpler for users to use a special `FunctionTerm` `fe()` to denote them.

src/contrasts.jl
nalimilan: OK. I was just wondering whether somebody actually needed to call this on `AbstractContrasts` objects without knowing whether they implement `levels` or not`. Returning an empty tuple would  be problematic as it could also mean "there are no levels".

src/contrasts.jl
nalimilan: OK. We could imagine returning a lazy matrix but I guess this isn't a bottleneck.

src/contrasts.jl
nalimilan: I don't think you addressed this, did you?

src/contrasts.jl
kleinschmidt: The only place where this is really used is in constructing the `ContrastsMatrix`.  All contrasts EVENTUALLY have to have levels, so whatever sentinel value we use is just to indicate that they haven't been manually specified ahead of time.

src/contrasts.jl
kleinschmidt: No, but since I'm ditching `Grouping` I thought it was okay :)

For the application I had in mind (random effect grouping terms for MixedModels.jl), there's really no utility in defining the order of the levels ahead of time, since it doesn't in any way affect the fitted model.  That's all that the levels stored on the contrasts are good for.

Also, this is called (IIRC) only after the `ContrastsMatrix` constructor has resolved any manually specified levels that are present on the contrasts with the levels that are present in the data, and that's what's being passed as the `levels` argument here.

kleinschmidt:
The only thing that's not hit by the existing tests is using an `AbstractTerm` as a hint directly, so I've added some tests for that (and in doing so discovered a bug...).  The `levels` methods are getting tested by the `ContrastsMatrix` tests which have pretty good coverage I think.

Issue #165, opened 2019-11-10
FormulaTerm type parameters


matthieugomez:
The FormulaTerm type is parametrized by the type of its terms. Is it really needed?
Because of this, a function such as `fit(x::MyModel, f::FormulaTerm)` is recompiled for each possible term types in the RHS or LHS. It sounds like this may lead to too much compilation.

oxinabox:
idk if it is needed or not.
I do agree that it can become problematic.
E.g. I know of people using StatsModels for large problems with say 3000 variables.
which makes for huge types.

Short-term: You could change definition of 
`fit(x::MyModel, f::FormulaTerm)` to `fit(x::MyModel, @nospecialize f::FormulaTerm)`
if that is a concern for you.

matthieugomez:
Thanks for the tip — this is what I do for now. Still, it may be better to simply remove these type parameters.

kleinschmidt:
That's a good point.  I'd originally put these type parameters on all the terms so that it would be possible to use `@generated` functions to produce very efficient methods for single-row `modelcols`, but I've not gotten around to that and it's probably a premature optimization.  I'm also now not sure that `@generated` functions are really the way to go for that, and moreover even if they are I think all you'd need is the "width" of each term...

Issue #147, opened 2019-09-02
parameters in formula interface


kkmann:
Hi,

I was just wondering whether someone considered adding support for parameterized functions to the formula interface. With the excellent automatic differentiation support in julia it should be possible to incorporate that in a genric fitting algorithm, right?

So, for instance, you could write something like

```julia
@formula(y ~ 1 + x + f(x, :alpha)
```
which would indicate that `:alpha` is a learnable parameter and needs to be fitted as well. Are there any technical/logical reasons why that shouldn't work?

PR #183, opened 2020-07-11
FunctionTerm is dead, long live FunctionTerm

kleinschmidt:
This is a pretty substantial change in how non-special function calls are
represented.  Instead of generating an anonymous function, the new
`FunctionTerm` simple wraps the called function, it's arguments (wrapped as
`Term`s), and the original expression.  When evaluated with `modelcols`, it uses
`Base.Broadcast.broadcasted` to lazily fuse nested function calls and then at
the top level calls `materialize` to run.  I hope that this will provide
performance that's comparable to the anonymous function, while being both more
run-time friendly and much simpler.

As a side effect, all of the macro-time parsing has been removed, and all the
special syntax now applies at run time (via method overloading of `+`, `&`, `~`,
and `*` (I know, I'm sorry, this is just for continuity's sake right now).

Also, ~~shamelessly stealing from~~ inspired by @oxinabox suggestions for how to
handle nested special and non-special functions (#117), I've added some
additional special syntax: `protect` and `unprotect`.  `protect` says to treat
every call below it as non-special, while `unprotect` (in an otherwise protected
context) says to treat everything below it as potentially special (e.g., as if
it occurred at the top level of a formula).  This isn't *perfect* at the moment
but it's a long way towards being able to do something like `1 - unprotect(poly(x, 3))` 
and having it do something sensible (actually that might work; what doesn't work
is `(1 - unprotect(a*b))` since that will generated `a + b + a&b` which don't
get fused into a single matrix which messes with the broadcasting).

I also added on a whim an `@support_unprotect op` macro which will designate a function as
special in unprotected contexts.  This simply adds a method like
``` julia
apply_schema(t::FunctionTerm{typeof($op)}, sch::Schema, Mod::Type) =
    apply_schema($op(t.args...), sch, Mod)
```
So, if you have something like `FunctionTerm{typeof(+)}` in a non-protected
context, then it will be converted into a call to `+(args[1], args[2], ...)`
when you do `apply_schema`.  This is taking advantage of the fact that all the
special handling of special syntax is done via method overloading of the
corresponding functions, so it should work fine at runtime.  At the moment it's
basically only useful for internal purposes but a two-argument form (function
and context type) might be useful for packages.

Tests do pass with this change, but it probably needs more tests to specifically
cover the protect/unprotect stuff.  ~~Docs aren't updated because that seems
premature at this point before folks have had a chance to weigh in.~~


kleinschmidt:
The major problem with this proposal, now that I've messed with it a bit, is that these kinds of methods really hit the compiler hard:

```julia
# + concatenates terms
Base.:+(terms::AbstractTerm...) = (unique(reduce(+, terms))..., )
Base.:+(a::AbstractTerm) = a
Base.:+(a::AbstractTerm, b::AbstractTerm) = (a,b)

# associative rule for +
Base.:+(as::TupleTerm, b::AbstractTerm) = (as..., b)
Base.:+(a::AbstractTerm, bs::TupleTerm) = (a, bs...)
Base.:+(as::TupleTerm, bs::TupleTerm) = (as..., bs...)
```

(`TupleTerm` is just `NTuple{AbstractTerm, N} where {N}`).

The reason is that a new method has to be compiled for every combination of number and types of terms being added together.  This is related to #165 and #147: all the type parameters in the term types create a lot of compiler overhead.  I can think of a few ways around this.  

1. we could go back to doing these transformations at parse time, which would require that `FunctionTerm` continue to hold onto two copies of the args (one parsed and the other not).  It would also require some duplication of functionality if we want to continue to support run-time construction as a first class interface, since we still need all these methods for the same syntax to apply as inside the macro.  

2. we could use some other non-parametrized version of the +-ed terms, like a Vector, or a custom wrapper.  This seems potentially the least disruptive (and could even convert back to a tuple if needed for performance at some point).

3. we could represnt ALL calls, even to special syntax like `+`, `&`, and `*`, as `FunctionTerm`s, and only parse them at `apply_schema` time, or do some fanciness during construction.  I think this is the most radical/involved and I'm not sure it's really worth it.

kleinschmidt:
Now I'm *really* confused.  I've actually gone through the exercise of replacing the tuple-based representation for `+`ed terms with a Vector-based storage (in [this branch](https://github.com/JuliaStats/StatsModels.jl/tree/dfk/vector)), and STILL get much longer first-call timings for every new number of terms, which suggests that some new methods are getting compiled, even though all the types are the same (no longer depend on the number of terms).

nalimilan:
`Base.:+(terms::AbstractTerm...)` is still going to be specialized on the number of arguments, right?

kleinschmidt:
Yeah, I'm afraid so.  Makes me wonder how stuff like this is handled in Base...

oxinabox:
Specialisation around splatting is weird in general, and its not documented what it will do.
IIRC `foo(x::VarArg)` and `foo(x...)` specialize differently.

But in anycase if you don't want specialiation just used the `@nospecialise` macro.

kleinschmidt:
I'm more and more concerned that just putting a few `@nospecialize` around the base methods we're overloading isn't going to be enough, since every time you hit a constructor for any term that can have children (formula, interaction, matrix, function) you're going to trigger compilation for every number and type of children, and same thing for every other method involving those types.  At that point you're going to have to put `@nospecialize` on just about everything that can take an AbstractTerm which seems like a sign that something has gone wrong.

At this point, my plan is to do a bit of benchmarking to see whether formula creation gets appreciably WORSE with this PR, and if not, we can forget about trying to fix the performance in another PR and focus on whether these changes do justice to @oxinabox vision enough to merge.

kleinschmidt:
Okay I've done a bit of timing (using the script from
https://gist.github.com/kleinschmidt/f51d305d56a590030c4f8688cbf18929).  On
master, these are the timings (first and second run times, since I'm mostly
interested in compilation time itself here):

``` julia
y ~ 1 + x
  0.000033 seconds (24 allocations: 1.453 KiB)
  0.000032 seconds (24 allocations: 1.453 KiB)
y ~ a + b
  0.008989 seconds (3.57 k allocations: 222.652 KiB)
  0.000014 seconds (11 allocations: 720 bytes)
y ~ a + b + c
  0.009176 seconds (3.58 k allocations: 223.059 KiB)
  0.000017 seconds (12 allocations: 752 bytes)
y ~ a + b + c + d
  0.011082 seconds (3.58 k allocations: 222.840 KiB)
  0.000017 seconds (13 allocations: 784 bytes)
y ~ a + b + c + d + e
  0.009841 seconds (3.58 k allocations: 223.012 KiB)
  0.000023 seconds (15 allocations: 896 bytes)
y ~ a + b + c + d + e + f
  0.010660 seconds (3.58 k allocations: 223.105 KiB)
  0.000029 seconds (16 allocations: 928 bytes)
y ~ log(a)
  0.065474 seconds (96.09 k allocations: 5.480 MiB)
  0.029364 seconds (9.05 k allocations: 520.255 KiB)
y ~ log(a) + log(b)
  0.941035 seconds (1.19 M allocations: 59.954 MiB, 10.87% gc time)
  0.746309 seconds (1.02 M allocations: 51.098 MiB, 12.56% gc time)
y ~ log(a) + log(b) + log(c)
  0.793438 seconds (1.43 M allocations: 72.068 MiB, 2.55% gc time)
  0.647390 seconds (901.92 k allocations: 45.123 MiB, 1.24% gc time)
y ~ log(a) + log(b) + log(c) + log(d)
  0.777564 seconds (1.01 M allocations: 50.078 MiB, 1.10% gc time)
  0.809190 seconds (1.01 M allocations: 50.081 MiB, 1.07% gc time)
y ~ log(a) + log(b) + log(c) + log(d) + log(e)
  0.914824 seconds (1.11 M allocations: 55.022 MiB, 1.15% gc time)
  0.918323 seconds (1.11 M allocations: 55.028 MiB, 1.02% gc time)
y ~ log(a) + log(b) + log(c) + log(d) + log(e) + log(f)
  1.036401 seconds (1.21 M allocations: 59.980 MiB, 1.71% gc time)
  1.017207 seconds (1.21 M allocations: 59.978 MiB, 0.86% gc time)
y ~ exp(a)
  0.027674 seconds (9.39 k allocations: 547.880 KiB)
  0.030770 seconds (9.05 k allocations: 520.177 KiB)
y ~ exp(a) + exp(b)
  0.778712 seconds (1.20 M allocations: 59.997 MiB, 2.29% gc time)
  0.684971 seconds (1.02 M allocations: 51.120 MiB, 1.34% gc time)
y ~ exp(a) + exp(b) + exp(c)
  0.742065 seconds (1.24 M allocations: 62.374 MiB, 1.24% gc time)
  0.655097 seconds (901.92 k allocations: 45.129 MiB, 1.56% gc time)
y ~ exp(a) + exp(b) + exp(c) + exp(d)
  0.773946 seconds (1.01 M allocations: 50.081 MiB, 1.32% gc time)
  0.778134 seconds (1.01 M allocations: 50.084 MiB, 2.41% gc time)
y ~ exp(a) + exp(b) + exp(c) + exp(d) + exp(e)
  0.891833 seconds (1.11 M allocations: 55.043 MiB, 1.04% gc time)
  0.873386 seconds (1.11 M allocations: 55.022 MiB, 1.05% gc time)
y ~ exp(a) + exp(b) + exp(c) + exp(d) + exp(e) + exp(f)
  0.982716 seconds (1.21 M allocations: 59.970 MiB, 1.11% gc time)
  1.017105 seconds (1.21 M allocations: 59.977 MiB, 1.86% gc time)
```

And on this PR (commit eaec084):

``` julia
y ~ 1 + x
  0.000100 seconds (25 allocations: 1.406 KiB)
  0.000093 seconds (25 allocations: 1.406 KiB)
y ~ a + b
  0.007148 seconds (2.17 k allocations: 135.030 KiB)
  0.000016 seconds (11 allocations: 768 bytes)
y ~ a + b + c
  0.042739 seconds (53.78 k allocations: 3.048 MiB)
  0.000026 seconds (23 allocations: 1.484 KiB)
y ~ a + b + c + d
  0.036280 seconds (53.91 k allocations: 3.052 MiB)
  0.000020 seconds (25 allocations: 1.547 KiB)
y ~ a + b + c + d + e
  0.047632 seconds (54.05 k allocations: 3.059 MiB)
  0.000022 seconds (28 allocations: 1.688 KiB)
y ~ a + b + c + d + e + f
  0.039089 seconds (54.18 k allocations: 3.064 MiB)
  0.000020 seconds (30 allocations: 1.750 KiB)
y ~ log(a)
  0.027760 seconds (44.74 k allocations: 2.699 MiB)
  0.000014 seconds (11 allocations: 480 bytes)
y ~ log(a) + log(b)
  0.277548 seconds (332.92 k allocations: 17.035 MiB)
  0.000028 seconds (27 allocations: 1.703 KiB)
y ~ log(a) + log(b) + log(c)
  0.080824 seconds (121.33 k allocations: 6.716 MiB)
  0.000031 seconds (42 allocations: 2.906 KiB)
y ~ log(a) + log(b) + log(c) + log(d)
  0.113681 seconds (114.74 k allocations: 6.305 MiB, 10.12% gc time)
  0.000034 seconds (49 allocations: 3.297 KiB)
y ~ log(a) + log(b) + log(c) + log(d) + log(e)
  0.075690 seconds (118.20 k allocations: 6.460 MiB)
  0.000034 seconds (57 allocations: 3.828 KiB)
y ~ log(a) + log(b) + log(c) + log(d) + log(e) + log(f)
  0.075736 seconds (121.95 k allocations: 6.627 MiB)
  0.000037 seconds (64 allocations: 4.219 KiB)
y ~ exp(a)
  0.013628 seconds (6.69 k allocations: 411.996 KiB)
  0.000013 seconds (11 allocations: 480 bytes)
y ~ exp(a) + exp(b)
  0.251576 seconds (332.96 k allocations: 17.036 MiB)
  0.000035 seconds (27 allocations: 1.703 KiB)
y ~ exp(a) + exp(b) + exp(c)
  0.083678 seconds (121.34 k allocations: 6.716 MiB)
  0.000037 seconds (42 allocations: 2.906 KiB)
y ~ exp(a) + exp(b) + exp(c) + exp(d)
  0.103966 seconds (114.75 k allocations: 6.304 MiB, 9.53% gc time)
  0.000035 seconds (49 allocations: 3.297 KiB)
y ~ exp(a) + exp(b) + exp(c) + exp(d) + exp(e)
  0.087459 seconds (118.21 k allocations: 6.461 MiB)
  0.000039 seconds (57 allocations: 3.828 KiB)
y ~ exp(a) + exp(b) + exp(c) + exp(d) + exp(e) + exp(f)
  0.080457 seconds (121.96 k allocations: 6.629 MiB)
  0.000038 seconds (64 allocations: 4.219 KiB)

```

Bottom line is, this PR is MUCH faster for anything involving a custom function,
both first and second run (noticeably so, going from ~1s every time you create a
formula with a function call in it to ~100ms first run and <1ms after that).
It's slower on first run (~50ms vs. ~10ms) but comparable after that.

There was, I remember, a good reason why it made sense to move the parsing rules
out of the macro and into run-time but at the moment I can't recall exactly
why.  Only that it had something to do with the `protect`/`unprotect` stuff.  I
don't _think_ it was just that I had started to dislike keeping both the
"parsed" and "non-parsed" copies of the arguments in the function term, but that
is a definite bonus in my view.

kleinschmidt:
I reverted the addition of test/Project.toml since that was the only thing breaking 1.0 compatibility, and am now using Compat for `only` instead of duplicating the code.

nalimilan:
I'm having a hard time trying to master the changes in this PR. :-)

Do you know why the new code is faster than the old one despite the cost of compilation? Maybe the compiler is smart enough to avoid compiling everything?

kleinschmidt:
There's two things going on.  Before, every `FunctionTerm` had a type parameter that was the type of the generated anonymous function.  A new anonymous function was generated every time, even if the call was the same.  So that means that `@formula(y ~ exp(a)` generates a `FunctionTerm{typeof(exp), typeof(SomeAnon)}`, and calling it again generates `FunctionTerm{typeof(exp), typeof(AnotherAnon)}`.  That's what's causing (I think) the slow times on master: any method with any argument that involves a function term is recompiled for every instance.

This PR fixes that, by removing the generated anonymous function altogether.  So each instance of `exp` is going to have the same type.  However, it also introduces the use of additional methods for `+`, `&`, and `*` operating on Terms to implement the algebraic rules that used to be handled at macro time.  This is necessary because of how the protect/unprotect logic (which is one of the main reasons for these changes to function terms) is implemented, as a function call.  See these lines: https://github.com/JuliaStats/StatsModels.jl/pull/183/files#diff-54c988da53bc06e26c2bc99378565e55R257-R289  The gist is that when you hit a `FunctionTerm{typeof(unprotect)}` during apply_schema, that means that you want to re-start applying the special parser rules (associative/distributive rules) below that point in the tree.  One way to do that is to implement those rules as methods for `&`, `+`, and `*` (that's these lines: https://github.com/JuliaStats/StatsModels.jl/pull/183/files#diff-13e0a130102f1c75064347b419d09b5fR375-R404), then when you hit a `FunctionTerm{typeof(+)}` etc. in an unprotected context, you can just do `term.f(term.args...)`, which is also (now) exactly what the `@formula` macro generates.

Beyond the advantages for simplifying the representation of function terms (by eliminating the need to keep protected and unprotected version of the arguments), this also has the benefit of making the DSL rules simpler (replacing 100-200 lines of code for the expr-based macro-time syntax rules with ~10-20 LOC for the methods) and the formula macro itself really minimal (all it does is wrap symbols as terms, and wrap non-special calls in function terms, which is also a much simpler operation now (this is it now: https://github.com/JuliaStats/StatsModels.jl/pull/183/files#diff-b4d7500de9148b52dc4251c7f5c66082R76-R84).

The downside is that there are more run-time calls which, given the current "everything is a type parameter" design, can be a bit rough on the compiler, since `y ~ a + b` has a different concrete type than `y ~ 1 + a` or `y ~ a + b + c` or `y ~ a + b + a&b`.  Not to mention that `+(a, b, c) has a different signature from `+(a, b, a&b)`, despite all three arguments being abstract terms.  I think these problems are fixable and there are good reasons to fix them anyway, but they do slow down formula construction a bit.  That was the point of the benchmarking above.  But in any case, I doubt that's the bottleneck in most applications, and it's a minimal price to pay given the massive speed ups we get whenever any function term is involved (as @matthieugomez reported in #164) at least on second run.

nalimilan:
Ah indeed, it's logical that not putting an anonymous function in the type parameter improves things.

src/terms.jl
```diff
@@ -475,13 +482,17 @@ julia> modelcols(MatrixTerm(ts), d)
...
-modelcols(t::Term, d::NamedTuple) =
-    throw(ArgumentError("can't generate modelcols for un-typed term $t. " *
-                        "Use apply_schema to create concrete terms first"))
+modelcols(t::Term, d::NamedTuple) = getproperty(d, t.sym)
+modelcols(t::ConstantTerm, d::NamedTuple) = t.n
+
+modelcols(ft::FunctionTerm, d::NamedTuple) =
+    Base.Broadcast.materialize(lazy_modelcols(ft, d))
+
+lazy_modelcols(ft::FunctionTerm, d::NamedTuple) =
```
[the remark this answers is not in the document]
kleinschmidt: see the line above...to get fusion with nested function terms the recursion happens with `lazy_modelcols` which returned a `Broadcasted`, and then `materialize` is called at the top level (the call to `modelcols`).

src/schema.jl
```diff
@@ -230,6 +232,63 @@ function apply_schema(t::ConstantTerm, schema::Schema, Mod::Type)
...
+
+
+# un-protecting special syntax: &, +, and *
+# 
+# if these occur as children of a non-special call, they are protected by
+# default.  so if we encounter them during apply_schema and we're NOT
+# in a Protected context, we need to call the corresponding op on the
+# # arguments of the FunctionCall
+
+# for op in (+, &, *)
```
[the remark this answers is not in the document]
kleinschmidt: So others can use it.  But I'm not 100% convinced it'd actually be useful...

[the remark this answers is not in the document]
kleinschmidt: Now that I've thought a bit more about this, I think it could be generally useful.  There's been some talk about developing better tooling for making extended syntax easier, and this seems like a step in that direction.

src/schema.jl
```diff
@@ -230,6 +232,63 @@ function apply_schema(t::ConstantTerm, schema::Schema, Mod::Type)
     InterceptTerm{t.n==1}()
 end
 
+# protection, borrowed from #117
```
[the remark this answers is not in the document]
kleinschmidt: Any reason why not?  I don't want to take total credit for @oxinabox idea :)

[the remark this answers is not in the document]
nalimilan: Well if we start putting references to issues and PR in the code, it's going to get very messy. `git blame` is here for this kind of thing. And #117 is mentioned in the OP, so it's fair!

kleinschmidt:
Another example encountered in the wild that is MASSIVELY improved by this PR:

```
@formula(y ~ 1 +  A * B * (C + C^2 + D) + (1 + A + B | G))
```

the issue is that before this PR, the `*` expansion is handled before anything else, so you end up with a ton of terms with `C^2`, and THEN `capture_call` is invoked, generating a DIFFERENT anonymous function for every individual term, which really hammers the compiler.

kleinschmidt:
This is ready for review again @nalimilan 

I think the failing project level codecov is due to deleting a lot of well-covered code so can be ignored (or rather is better fixed in a separate PR which adds more tests) since the patch coverage is high.

src/schema.jl
```diff
@@ -237,6 +239,150 @@ function apply_schema(t::ConstantTerm, schema::Schema, Mod::Type)
...
+context is when a call to [`unprotect`](@ref) is encountered.  At that point,
+everything below the call to `unprotect` is treated as special formula syntax.
+
+A `Protected` context is created inside a [`FunctionTerm`](@ref) automatically,
+but can be manually created with a call to [`protect`](@ref).
+```
+"""
+struct Protected{Ctx} end
+Base.broadcastable(x::Protected) = Ref(x)
+"""
```
nalimilan, line 267:
```suggestion
"""

```

src/schema.jl
```diff
@@ -237,6 +239,150 @@ function apply_schema(t::ConstantTerm, schema::Schema, Mod::Type)
...
+julia> d.a .+ d.b
+4-element Array{Float64,1}:
+ 1.4886128300795012
+ 2.210968202158536
+ 3.951916339835734
+ 4.999904658898614
+```
+"""
+protect(ctx) = Protected{ctx}()
+# construct singletons to avoid method ambiguities using Type{<:Protected}
```
nalimilan, line 298: I don't understand to what code this comment applies.

src/schema.jl
```diff
@@ -237,6 +239,150 @@ function apply_schema(t::ConstantTerm, schema::Schema, Mod::Type)
...
+    tt = only(t.args)
+    apply_schema(tt, schema, Ctx)
+end
+
+"""
+    unprotect(term)
+    unprotect(::Protected{T})
+
+Inside a [`@formula`], removes [`Protected`](@ref) status for the argument
+term(s).  This allows the usual special [`@formula`](@ref) interpretation of
```
nalimilan, line 325: Same remark about "usual special".

src/schema.jl
```diff
@@ -237,6 +239,150 @@ function apply_schema(t::ConstantTerm, schema::Schema, Mod::Type)
...
+
+julia> 1 .- d.a .* d.b
+4×1 Array{Float64,2}:
+  0.5113871699204988
+  0.5780635956829281
+ -1.855749019507202
+ -2.9996186355944543
+```
+"""
+unprotect(::Protected{Ctx}) where Ctx = Ctx
```
nalimilan, line 353:
```suggestion
unprotect(::Protected{Ctx}) where {Ctx} = Ctx
```

src/schema.jl
```diff
@@ -237,6 +239,150 @@ function apply_schema(t::ConstantTerm, schema::Schema, Mod::Type)
...
+
+function apply_schema(t::FunctionTerm{typeof(unprotect)}, schema::Schema, Ctx::Protected)
+    tt = only(t.args)
+    apply_schema(tt, schema, unprotect(Ctx))
+end
+
+# un-protecting special syntax: &, +, and *
+# 
+# if these occur as children of a non-special call, they are protected by
```
nalimilan, line 362: Here "special" seems to be used to mean "normal Julia syntax." :-)

src/schema.jl
```diff
@@ -237,6 +239,150 @@ function apply_schema(t::ConstantTerm, schema::Schema, Mod::Type)
...
+# default.  so if we encounter them during apply_schema and we're NOT
+# in a Protected context, we need to call the corresponding op on the
+# # arguments of the FunctionCall
```
nalimilan, line 365:
```suggestion
# arguments of the FunctionCall
```

src/schema.jl
```diff
@@ -237,6 +239,150 @@ function apply_schema(t::ConstantTerm, schema::Schema, Mod::Type)
...
+# macro unprotect(op)
+#     esc(quote
+#         apply_schema(t::StatsModels.FunctionTerm{typeof($op)}, sch::StatsModels.Schema, Mod::Type) =
+#             apply_schema(t.f(t.args...), sch, Mod)
+#     end)
+# end
+
+# @unprotect(+)
+# @unprotect(&)
+# @unprotect(*)
```
nalimilan, line 383: As you prefer, but either use it or remove it?

src/temporal_terms.jl
```diff
@@ -30,11 +30,11 @@ terms(t::LeadLagTerm) = (t.term, )
...
     opname = string(nameof(F.instance))
-    if length(t.args_parsed) == 1  # lag(term)
-        term_parsed = first(t.args_parsed)
+    if length(t.args) == 1  # lag(term)
+        term_parsed = first(t.args)
         nsteps = 1
-    elseif length(t.args_parsed) == 2  # lag(term, nsteps)
-        term_parsed, nsteps_parsed = t.args_parsed
+    elseif length(t.args) == 2  # lag(term, nsteps)
+        term_parsed, nsteps_parsed = t.args
```
nalimilan, line 37: Drop the "parsed" terminology?

src/terms.jl
```diff
@@ -116,18 +118,14 @@ julia> modelcols(f.rhs, (a=[3, 4], b=[4, 5]))
...
 """
-struct FunctionTerm{Forig,Fanon,Names} <: AbstractTerm
-    forig::Forig
-    fanon::Fanon
+struct FunctionTerm{F,Args} <: AbstractTerm
+    f::F
+    args::Args
     exorig::Expr
-    args_parsed::Vector
 end
```
nalimilan, line 125: Is it actually useful to parameterize on the function and argument types? Doesn't `broadcast` specialize on these anyway?

src/terms.jl
```diff
@@ -398,18 +376,52 @@ Base.show(io::IO, mime::MIME"text/plain", t::MatrixTerm; prefix="") =
...
+
+# associative rule for +
 Base.:+(as::TupleTerm, b::AbstractTerm) = b in as ? as : (as..., b)
 Base.:+(a::AbstractTerm, bs::TupleTerm) = a in bs ? bs : (a, bs...)
 Base.:+(as::TupleTerm, bs::TupleTerm) = (union(as, bs)..., )
 
+# * expansion
+Base.:*(a::TermOrTerms, b::TermOrTerms) = a + b + a&b
+
+cleanup(terms::TupleTerm) = tuple(sort!(unique!([terms...]), by=degree)...)
```
nalimilan, line 417:
```suggestion
cleanup(terms::TupleTerm) = Tuple(sort!(unique!(collect(terms)), by=degree))
```

src/terms.jl
```diff
@@ -398,18 +376,52 @@ Base.show(io::IO, mime::MIME"text/plain", t::MatrixTerm; prefix="") =
...
+cleanup(x) = x
+
+degree(::AbstractTerm) = 1
+degree(t::InteractionTerm) = mapreduce(degree, +, t.terms)
+# dirty hack, move to MixedModels.jl
+degree(::FunctionTerm{typeof(|)}) = Inf
```
nalimilan, line 423: This means that this PR will actually be breaking if any packages define similar terms?

test/extension.jl
```diff
@@ -8,9 +8,10 @@ struct PolyTerm <: AbstractTerm
     deg::Int
 end
 PolyTerm(t::Term, deg::ConstantTerm) = PolyTerm(t.sym, deg.n)
+poly(t::AbstractTerm, deg) = PolyTerm(t, deg)
 
-StatsModels.apply_schema(t::FunctionTerm{typeof(poly)}, sch, ::Type{<:PolyModel}) =
-    PolyTerm(t.args_parsed...)
+StatsModels.apply_schema(t::FunctionTerm{typeof(poly)}, sch::Schema, Mod::Type{<:PolyModel}) =
+    apply_schema(poly(t.args...), sch, Mod)
```
nalimilan, line 14: Here too, what's going to happen if a package uses a similar pattern to define a custom term?

test/extension.jl
```diff
@@ -36,7 +37,8 @@ end
 
         f_plain = apply_schema(f, sch)
         @test f_plain.rhs.terms[1] isa FunctionTerm
-        @test f_plain == apply_schema(f, sch, Nothing)
+        # this is true but == not defined correctly and apply_schema creates a new instance
+        @test_broken f_plain == apply_schema(f, sch, Nothing)
```
nalimilan, line 41: What does "this is true" means?

test/protect.jl
```diff
@@ -0,0 +1,43 @@
...
+
+        # ideally you'd also be able to do these but it's hard to make it work...
+        # this fails because - doesn't auto-broadcast, and the broadcasting that
+        # happens during FunctionTerm evaluation gets used up by the (a,b) tuple.
+        ff = apply_schema(@formula(0 ~ 1 - unprotect(a+b)), schema(d))
+        @test_broken modelcols(ff.rhs, d) == 1 .- [d.a d.b]
+
+        # and even if we define a broadcasting version, still fails because of
+        # packing the result into a right-sized container.  end up with a tuple
+        # of arrays which needs to be collected into a matrix
```
nalimilan, line 34:
```suggestion
        # and even if we define a broadcasting version, still fails because it
        # gives a tuple of arrays instead of a matrix
```

test/protect.jl
```diff
@@ -0,0 +1,43 @@
...
+        my_sub = (x,y) -> x .- y
+        ff = apply_schema(@formula(0 ~ my_sub(1, unprotect(a+b))), schema(d))
+        @test_broken modelcols(ff.rhs, d) == 1 .- [d.a d.b]
+        
+        # both of these could be fixed by always returning a matrix when you call
+        # modelcols on a tuple of terms but that would break other things
```
nalimilan, line 40: Do you think packages may depend on this?

src/terms.jl
```diff
@@ -398,18 +376,52 @@ Base.show(io::IO, mime::MIME"text/plain", t::MatrixTerm; prefix="") =
...
-
+Base.:&(a::AbstractTerm, b::AbstractTerm) = InteractionTerm((a,b))
+
+Base.:&(::ConstantTerm, b::AbstractTerm) = b
+Base.:&(a::AbstractTerm, ::ConstantTerm) = a
+
+# Avoid method ambiguities
+Base.:&(::ConstantTerm, b::InteractionTerm) = b
+Base.:&(a::InteractionTerm, ::ConstantTerm) = a
+Base.:&(a::ConstantTerm, ::ConstantTerm) = a
```
nalimilan, line 390: `ConstantTerm` allows storing a constant value, which is completely ignored here. Is this intended?

nalimilan (commented):
Sorry for the super long delay, for some reason I thought this wasn't ready. You should feel free to ping me every year or so... :-p

A handful of new lines don't seem to be covered by tests. Are these real?

src/terms.jl
kleinschmidt: mostly laziness on my part; we could add a check that it stores a `1` and throw an error otherwise?

test/protect.jl
kleinschmidt: yes, at least MixedModels.jl does!  it relies on the distinction between ranef terms being "non matrix terms" and fixef terms being "matrix terms" to create a tuple of `(allfixefterms_matrix, ranef1_matrix, ranef2_matrix, ...)`

kleinschmidt: MAYBE we coudl work around that by using some kind of "lazy model matrix" system where you don't actually collect everything into a `Matrix`

test/extension.jl
kleinschmidt: they are the same under a reasonable informal definition of "equal" but `==` isn't defined properly (as far as I can remember, it's been some time :joy: )

docs/src/api.md
```diff
@@ -48,6 +48,16 @@ collect_matrix_terms
 is_matrix_term
 ```
 
+### Protection
+
+For controlling when
```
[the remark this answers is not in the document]
kleinschmidt: I mean, you get the idea right

src/schema.jl
kleinschmidt: the bit above.  `protect(ctx)` _could_ return `Protected{ctx}` but that leads to ambiguities using `Type{<:Protected}` for dispatch instaed of `Protected` for some reason I've now forgotten

src/terms.jl
kleinschmidt: we need the function at least for methods like `apply_schema` to handle specific syntax.  the args I could see the case for dropping sicne they go through the function barrier in broadcast, but I'd prefer to make that change uniformly for all teh term types to see what the overall runtime and compile time performance implications are

src/terms.jl
kleinschmidt: "similar terms" meaning using `typeof(|)` for special syntax?

test/extension.jl
kleinschmidt: same as always, you'll get method invalidation errors and/or the context will shuttle dispatch in the right way in the right model contexts

kleinschmidt: method invalidation *warnings, not errors

src/schema.jl
nalimilan: OK. Then maybe this would be more explicit?
```suggestion
# return instances rather than types to avoid method ambiguities using Type{<:Protected}
```

src/terms.jl
nalimilan: No, since apparently this PR requires changing something in MixedModels or adding a hack here, could it affect other packages too?

test/extension.jl
nalimilan: OK. Then maybe this? :-)
```suggestion
        # this works but == is not defined correctly and apply_schema creates a new instance
        @test_broken f_plain == apply_schema(f, sch, Nothing)
```

src/terms.jl
nalimilan: Why not avoid storing anything in it then?

test/protect.jl
nalimilan: I'm a bit lost... What's the relationship between this and `unprotect`? Is the concern only that it's breaking, or also that the API isn't completely satisfying?

src/terms.jl
kleinschmidt: Because it's what constant numeric values get lowered to during `@formula`, so we need to be able to access them for things like `@formula(poly(a, 3))`.

src/terms.jl
kleinschmidt: huh, TIL!

src/terms.jl
kleinschmidt: ah yeah I see now.  we are already special casing calls to `:|` so this is not breaking:

https://github.com/JuliaStats/StatsModels.jl/blob/463eb0acb49bc5428374d749c4da90ea2a6c74c4/src/formula.jl#L272-L274

test/protect.jl
kleinschmidt: The main concern is that it's breaking; the most salient case off the top of my head is that MixedModels uses `extract_matrix_terms` to separate out the fixed effect terms from the random effects terms, and then calls `modelcols` on teh tuple of terms that results.

src/schema.jl
kleinschmidt: okay, having thought about this a bit more, I think this _is_ useful since it essentially implements the recommended pattern for custom syntax now (to make methods for the function that operate on terms and return some kind of `AbstractTerm` that we can operate on with `apply_schema`, `modlecols`, etc.).

src/terms.jl
kleinschmidt: (about `Tuple` constructor taking an iterable...)

src/schema.jl
kleinschmidt: Nope, still means `+`, `&` etc.  The situation this is describing is something like

```
protect(unprotect(a * b))
```

or

```
log(unprotect(a & b))
# -> FunctionTerm(log, [InteractionTerm(Term(:a), Term(:b))], ...)
```

(since any non-special call implicitly protects it's children).  So in this case, teh call to `&` occurs as a child to a call to `log`, but in a non-protected context.

docs/src/api.md
```diff
@@ -50,11 +50,14 @@ is_matrix_term
 
 ### Protection
 
-For controlling when 
+For more fine-grained control over whether function calls are treated as normal
+julia calls ("protected" and captured as `FunctionTerm`s) or as `@formula`
```
nalimilan, line 54:
```suggestion
Julia calls ("protected" and captured as `FunctionTerm`s) or as `@formula`
```

src/schema.jl
```diff
@@ -366,37 +368,67 @@ julia> 1 .- d.a .* d.b
...
-# if these occur as children of a non-special call, they are protected by
-# default.  so if we encounter them during apply_schema and we're NOT
-# in a Protected context, we need to call the corresponding op on the
-# # arguments of the FunctionCall
-for op in (+, &, *)
-    @eval begin
-        apply_schema(t::FunctionTerm{typeof($op)}, sch::Schema, Mod::Type) =
+"""
+    StatsModels.@unprotect f
+
```
nalimilan, line 380: `@unprotect` may not be the best name, as usually `@func x` is a convenience equivalent of `func(x)`, which isn't the case here. Maybe something like `@support_unprotect` or even `@special_term`?

Also:
```suggestion
    StatsModels.@unprotect f

Generate methods necessary for function `f` to support [`unprotect`](@ref).

```

Project.toml
```diff
@@ -15,6 +16,7 @@ Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 
 [compat]
 CategoricalArrays = "0.8"
+Compat = "2.2, 3"
```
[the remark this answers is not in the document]
nalimilan: This line isn't in the diff anymore. Is this intended?

src/terms.jl
nalimilan: OK. It would sound safer to check that the value is 1 then.

test/protect.jl
nalimilan: Sorry, don't have a clear view of all this. Anyway I guess the two main questions are: does this PR break MixedModels? does it allow fixing this later without breaking the API?

src/schema.jl
kleinschmidt: yeah that's a good point, I hadn't thought about how there was already a function called `unprotect`.  I think `@support_unprotect` is a good name.

src/terms.jl
kleinschmidt: The current behavior on 0.6.x is a bit inconsistent here between `@formula` and run-time:

```julia
julia> @formula(y ~ 1 & 2)
┌ Warning: Number 2 removed from interaction term (&)(2)
└ @ StatsModels ~/.julia/packages/StatsModels/fK0P3/src/formula.jl:158
FormulaTerm
Response:
  y(unknown)
Predictors:

julia> term(1) & term(2)
1 & 2
```

So whatever we do here, we're going to be inconsistent with ONE of the two paths lol.

As it currently stands in this PR, we're closer to what happens in `@formula` on main: constant terms are removed from interactions regardless of their value.  Should we be making it an error to have a constant that's not `1` in an interaction?  That seems a bit aggressive to me but I also have a hard time thinking of cases when that would actually come up, so making it an error might not be that bad of an idea since this is a breaking change anyway, and it makes it possible to change the behavior later with less risk that someone is relying on essentially undefined behavior...

Project.toml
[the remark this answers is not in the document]
kleinschmidt: yup, Compat was only used for `only` pre-1.6 and we now support 1.6+ (since #252 )

Project.toml
```diff
@@ -1,8 +1,9 @@
 name = "StatsModels"
 uuid = "3eaba693-59b7-5ba5-a881-562e759f1c8d"
-version = "0.6.33"
+version = "0.7.0-DEV"
```
nalimilan, line 3: Usually we just set the version that will be released (I'm not even sure Pkg accepts letters, which may explain the docs build failure).

```suggestion
version = "0.7.0"
```

src/terms.jl
```diff
@@ -55,38 +57,25 @@ struct FormulaTerm{L,R} <: AbstractTerm
...
-of the function as originally specified (e.g., `typeof(log)`), while the second
-is the type of the anonymous function that will be applied element-wise to the
-data table.
+    FunctionTerm{F,Args} <: AbstractTerm
 
-The `FunctionTerm` _also_ captures the arguments of the original call and parses
-them _as if_ they were part of a special DSL call, applying the rules to expand
-`*`, distribute `&` over `+`, and wrap symbols in `Term`s.
+Represents a call to a Julia function.  The first type parameter is the type of
+the captured function (e.g., `typeof(log)`), and the second is the types of the
```
nalimilan, line 63:
```suggestion
the captured function (e.g., `typeof(log)`), and the second is the type of the
```

src/terms.jl
nalimilan: Yes better throw errors in absurd cases rather than risk hiding bugs. So with the new state of the PR an error is thrown in both cases above, right?

nalimilan (approved):

Project.toml
nalimilan: Actually we should even do this:
```suggestion
version = "1.0.0"
```

src/terms.jl
kleinschmidt: yup, any time a ConstantTerm is involved in a `&` we check that it's `1` and throw an error otherwise!

PR #188, opened 2020-08-28
Bump CategoricalArrays to 0.8

nalimilan:
We claim to support DataFrames 0.21, but it wasn't actually used since we didn't support CategoricalArrays 0.8.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -14,9 +14,9 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 
 [compat]
-CategoricalArrays = "0.7"
+CategoricalArrays = "0.8"
 DataAPI = "1.1"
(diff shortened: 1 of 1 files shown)


kleinschmidt:
I don't think a new release is necessary after this since installing 0.6.13 with DataFrames already gives you 0.21 (and 0.8 of categorical arrays)

nalimilan:
Ah, right, because these are only test dependencies. That's kind of a silly behavior though: users are allowed to use these versions, but we're not allowed to test them. :-/

kleinschmidt:
Yeah I guess that's why we moved to having test dependencies in test/Project.toml (more clearly separate the requirements for testing and using), but that's not supported in 1.0 so we'll have to wait until next LTS to do that.

nalimilan:
AFAICT that doesn't change anything to the fact that users get different versions than our tests...

kleinschmidt:
it's at least clearer that the test dependencies have nothing to do with the installed dependencies though, which AFAIK is the situation for anything in `[targets] test=...`

PR #190, opened 2020-08-29
Add string method for `term`

nilshg:
As discussed on Slack, this makes it easier to create `Term` objects from `DataFrame` columns, as `names(df)` now returns strings.

### DIFF
--- a/docs/src/formula.md
+++ b/docs/src/formula.md
@@ -299,10 +299,10 @@ Predictors:
     long as the packages have [implemented support for them](@ref extend-runtime).
 
 The [`term`](@ref) function constructs a term of the appropriate type from
-symbols and numbers, which makes it easy to work with collections of mixed type:
+symbols, strings and numbers, which makes it easy to work with collections of mixed type:
 
 ```jldoctest 1
-julia> ts = term.((1, :a, :b))
+julia> ts = term.((1, :a, "b"))
 1
 a(unknown)
 b(unknown)
--- a/docs/src/formula.md
+++ b/docs/src/formula.md
@@ -299,7 +299,8 @@ Predictors:
     long as the packages have [implemented support for them](@ref extend-runtime).
 
 The [`term`](@ref) function constructs a term of the appropriate type from
-symbols, strings and numbers, which makes it easy to work with collections of mixed type:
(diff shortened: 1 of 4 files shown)


kleinschmidt:
Looks great. Can you add something to the docs/src/formula.md file too?  `term` is described there as well.

kleinschmidt:
Oh and add some tests!

nilshg:
> Oh and add some tests!

Ah right I was wondering about that - I thought the `jldoctest` stuff is a test, so that changing the example in the docstring to include a string would be the test. Is that not actually the case or is it just that doctests are additional to normal tests, not a replacement?

kleinschmidt:
Unfortunately not (you can see where they run in the documentation job below...rather than the travis job).  We currently duplicate many of them in the actual tests.  They're mostly used to make sure there aren't broken examples in the docs :)

nilshg:
Pretty bare bones but I think this might do the trick?

docs/src/formula.md
```diff
@@ -299,10 +299,10 @@ Predictors:
     long as the packages have [implemented support for them](@ref extend-runtime).
 
 The [`term`](@ref) function constructs a term of the appropriate type from
-symbols and numbers, which makes it easy to work with collections of mixed type:
+symbols, strings and numbers, which makes it easy to work with collections of mixed type:
```
kleinschmidt, line 302: Maybe just to be a bit more explicit (@nalimilan opinions?)

```suggestion
symbols or strings (`Term`) and numbers (`ConstantTerm`), which makes it easy to 
work with collections of mixed type:
```

src/terms.jl
```diff
@@ -580,12 +580,13 @@ hasresponse(t::FormulaTerm) =
     term(x)
 
 Wrap argument in an appropriate `AbstractTerm` type: `Symbol`s become `Term`s,
```
nalimilan, line 582:
```suggestion
Wrap argument in an appropriate `AbstractTerm` type: `Symbol`s and `AbstractString`s become `Term`s,
```
(Needs a line break but you get the idea.)

kleinschmidt:
LGTM!  Last thing: bump the patch version in Project.toml?  Thanks for the contribution!

PR #192, opened 2020-09-15
fix doctests and add missing docstrings

kleinschmidt:
This fixes up the doctests a bit, which had gotten out of sync with the actual
package and with how GLM prints coef tables.  The most substantive changes are
to add some docstrings that were missing (`response` and `modelmatrix`) and to
update the docstring and examples for `hypothesis_matrix` to reflect the fact
that this now directly returns the pseudo-inverse (where the interpretation of
rows and columns is switched from the contrasts matrix).


kleinschmidt:
I'll rebase and merge this if there are no objections.

docs/src/formula.md
```diff
@@ -366,18 +366,18 @@ StatsModels.TableRegressionModel{LinearModel{GLM.LmResp{Array{Float64,1}},GLM.De
...
+───────────────────────────────────────────────────────────────────────
+(Intercept)  1.01518   0.0378035  26.85    <1e-44   0.940101    1.09026
+a            1.98369   0.0766599  25.88    <1e-43   1.83144     2.13595
+b: e         3.03157   0.0606258  50.00    <1e-67   2.91116     3.15198
+b: f         3.98553   0.0563457  70.73    <1e-81   3.87362     4.09744
+b: g         5.00412   0.0577589  86.64    <1e-89   4.88941     5.11884
+a & b: e     5.91659   0.105766   55.94    <1e-72   5.70653     6.12665
+a & b: f     6.98546   0.110637   63.14    <1e-76   6.76573     7.2052
+a & b: g     7.95902   0.109459   72.71    <1e-82   7.74163     8.17642
+───────────────────────────────────────────────────────────────────────
```
palday, line 380: Not just the rounding but the actual values changed here. Is that correct?

docs/src/internals.md
```diff
@@ -518,15 +518,15 @@ StatsModels.TableRegressionModel{LinearModel{GLM.LmResp{Array{Float64,1}},GLM.De
...
-b^2           2.89786    0.0794572  36.4707     <1e-56   2.74012     3.05561
-────────────────────────────────────────────────────────────────────────────
+─────────────────────────────────────────────────────────────────────────
+                 Coef.  Std. Error      t  Pr(>|t|)  Lower 95%  Upper 95%
+─────────────────────────────────────────────────────────────────────────
+(Intercept)   0.642922   0.154789    4.15    <1e-4    0.335627   0.950218
+a^1           1.88667    0.382831    4.93    <1e-5    1.12665    2.64668
+a^2           2.57741    1.37558     1.87    0.0640  -0.15346    5.30828
+b^1          -0.22193    0.148416   -1.50    0.1381  -0.516573   0.072714
+b^2           2.88983    0.0790131  36.57    <1e-57   2.73297    3.04669
```
palday, line 528: likewise here

docs/src/internals.md
```diff
@@ -627,15 +627,15 @@ StatsModels.TableRegressionModel{LinearModel{GLM.LmResp{Array{Float64,1}},GLM.De
...
+─────────────────────────────────────────────────────────────────────────
```
palday, line 638: and here

docs/src/internals.md
```diff
@@ -734,26 +734,26 @@ StatsModels.TableRegressionModel{LinearModel{GLM.LmResp{Array{Float64,1}},GLM.De
...
-───────────────────────────────────────────────────────────────────────────
-(Intercept)  0.911363    0.310486   2.93528    0.0042   0.295214    1.52751
-poly(b, 2)   2.94442     0.191024  15.4139     <1e-27   2.56534     3.3235
-───────────────────────────────────────────────────────────────────────────
+────────────────────────────────────────────────────────────────────────
+                Coef.  Std. Error      t  Pr(>|t|)  Lower 95%  Upper 95%
+────────────────────────────────────────────────────────────────────────
+(Intercept)  0.933658    0.286572   3.26    0.0015   0.364965    1.50235
+poly(b, 2)   2.9377      0.179058  16.41    <1e-29   2.58237     3.29304
+────────────────────────────────────────────────────────────────────────
```
palday, line 742: and here, although less extreme

docs/src/internals.md
```diff
@@ -734,26 +734,26 @@ StatsModels.TableRegressionModel{LinearModel{GLM.LmResp{Array{Float64,1}},GLM.De
...
-b^1          2.13096    0.100552   21.1926    <1e-98   1.93388     2.32804
-b^2          3.1132     0.0813107  38.2877    <1e-99   2.95384     3.27257
-──────────────────────────────────────────────────────────────────────────
+────────────────────────────────────────────────────────────────────────
+                Coef.  Std. Error      z  Pr(>|z|)  Lower 95%  Upper 95%
+────────────────────────────────────────────────────────────────────────
+(Intercept)  0.843637   0.131701    6.41    <1e-9    0.585508    1.10177
+b^1          1.95267    0.101847   19.17    <1e-81   1.75305     2.15229
+b^2          3.05973    0.0824839  37.09    <1e-99   2.89806     3.22139
+────────────────────────────────────────────────────────────────────────
```
palday, line 756: another one

src/contrasts.jl
```diff
@@ -660,6 +662,11 @@ non-zero mean).  If `tolerance != 0` (the default), the hypotheses are rounded
 to `Int`s if possible and `Rational`s if not, using the given tolerance.  If
 `tolerance == 0`, then the hypothesis matrix is returned as-is.
 
+The orientation of the hypothesis matrix is _opposite_ that of the contrast
+matrix: each row of the contrasts matrix is a data level and each column is a
+predictor, whereas each row of the hypothesis matrix is the interpretation of a
+predictor with weights for each level given in the columns.
+
 Note that this assumes a *balanced design* where there are the same number of
```
palday, line 670: Yep, super important to make this clear. :+1:

palday (commented):
Not just the rounding but the values of a lot of the coefficient tables changed. Was this intentional?

kleinschmidt:
I think the different coefficients are due to the docs being built on 1.5 instead of 1.3 previously and relying on RNG...

palday:
Ah yeah, the RNG changes are a pain.

palday:
Can you change the RNG to use `StableRNGs`? Then we won't have these changes in the future. Either way, go ahead and rebase.

kleinschmidt:
I thought about that but it would require tweaking the set-up too and for some reason didn't seem worth it, although now I can't quite remember why.  I'll have another look...

kleinschmidt:
Ah I remember what it was: I'd have to also update the doctests to also take the `rng` argument.  That's not a huge deal I guess especially since it only touches the data generation part.

palday:
I didn't comment on it, but I was curious why you set the seed in the setup because then everything else depends on it always being executed in that order and with no intervening calls to the RNG.

kleinschmidt:
The setup block runs for every doc test (except when there's a labeled block in which case the order does matter)

palday:
Ah, good to know!

palday (approved):

palday:
Still got a merge conflict but otherwise LGTM

kleinschmidt:
I still gotta fix the doctests...just wanted to run CI and see what happened here :)

kleinschmidt:
Okay, a few doctests do fail, but they're differences in the path between the builder and my local path, and some strange differences in very small numbers that are only printed to illustrate that they're basically zero and are therefore distracting...

palday:
@ararslan or @nalimilan should comment on how important they find such things.

kleinschmidt:
yeah it would be a matter of adding a regex to ignore those failures...I don't know why the smol numbers are different given that it's deterministic input and just a matrix pseudoinverse.

palday:
Let me tell the horrors of platform-dependence numerical linear algebra and the holy mess that incurred in rank deficiency detection in MixedModels ..... 

::wimpers::

ararslan:
> @ararslan or @nalimilan should comment on how important they find such things.

This is literally the only thing that matters and this is a hill I'm willing to die on.

(Actually I don't care at all)

palday:
I don't really care per se, but I do know that if you get used to ignoring an 'alarm'/warning, you wind up ignoring it when you shouldn't. The bot form for the boy who cried wolf.

ararslan:
I would suggest that any doctests that are susceptible to numerical funny business either be converted from `jldoctest` to `julia-repl` or weaseled such that some transformation is applied as the last step that gets displayed.

kleinschmidt:
Here's the contents of that doctest:

````
```jldoctest
julia> cmat = StatsModels.contrasts_matrix(DummyCoding(), 1, 4)
4×3 Array{Float64,2}:
 0.0  0.0  0.0
 1.0  0.0  0.0
 0.0  1.0  0.0
 0.0  0.0  1.0

julia> StatsModels.hypothesis_matrix(cmat)
4×4 Array{Int64,2}:
  1  0  0  0
 -1  1  0  0
 -1  0  1  0
 -1  0  0  1

julia> StatsModels.hypothesis_matrix(cmat, intercept=false) # wrong without intercept!!
3×4 Array{Int64,2}:
 0  1  0  0
 0  0  1  0
 0  0  0  1

julia> StatsModels.hypothesis_matrix(cmat, tolerance=0) # ugly
4×4 Array{Float64,2}:
  1.0  -2.23753e-16   6.91749e-18  -1.31485e-16
 -1.0   1.0          -2.42066e-16   9.93754e-17
 -1.0   4.94472e-17   1.0           9.93754e-17
 -1.0   1.04958e-16  -1.31044e-16   1.0        

julia> StatsModels.hypothesis_matrix(StatsModels.ContrastsMatrix(DummyCoding(), ["a", "b", "c", "d"]))
4×4 Array{Int64,2}:
  1  0  0  0
 -1  1  0  0
 -1  0  1  0
 -1  0  0  1

```
````

kleinschmidt:
cf. `# ugly` on that line...

ararslan:
Could have a doctest dependency on ZChop and just call `zchop` to force the tiny bits to zero, or just like `round.(Int, _)`...

kleinschmidt:
That's basically what the `tolerance` kwarg does (clips things to `Int` or `Rational`).  The point of the example is to show what happens when you set it to zero (just passes the floats through as is).

palday:
@kleinschmidt Maybe show that it's not exactly (elementwise) equal to 0, ±1, but that it is less than an epsilon away from that.

nalimilan:
Sounds good. Though doctests are only run on a given Julia version so we could also hardcode whatever result we obtain there, right?

kleinschmidt:
I thought I _was_ running it on the same version locally (1.5.3) but maybe not, or maybe it depends somehow on the architecture it's running on, who knows.  I think replacing the `jldoctest` with `juliarepl` block is a good solution.

kleinschmidt:
Oh good the printing of type parameters (now there are spaces printed after each comma) and matrix/vector types (`Matrix{T}` instead of `Array{T,2}`) is now different on nightly so a million doctests are broken again.  I think at this point I should just give up supporting both 1.6 and 1.5 in the doctests and make another fix once 1.6 is actually released.

PR #194, opened 2020-09-21
Switch integration to travis-ci.com

kleinschmidt:
Only change is to badge at this point; migration is done on the Travis end
already.


kleinschmidt:
This also required removing the travis webhook and turn the github app integration for this repository off and on again.

PR #195, opened 2020-09-30
dep from StatsFuns instead of Distributions

matthieugomez:
`Distribution` depends  on `FillArrays`, `StaticArrays`, etc, that create invalidations. On my computer, removing this dependency decreases `using StatsModels` by a third (after precompilation)


palday (approved):
LGTM, I'd wait for @kleinschmidt or @nalimilan to double check.

ararslan:
Looks like it's only being used for p-values for the likelihood ratio test. Shouldn't that code live in HypothesisTests?

palday:
Just when we got `lrtest` merged .... :stuck_out_tongue_winking_eye:

matthieugomez:
Yes I’m not even sure why this code is here.

nalimilan (approved):
Fine with me. Though the idea was that we would create a StatsModelsBase package if people wanted a more lightweight dependency.

Regarding the choice between HypothesisTests and StatsModels, I think it makes more sense to put modelling-related functions in StatsModels, as we will be able to extract and print formulas describing each model in the table (currently not done because of the lack of generic API). HypothesisTests doesn't know about formulas at all. More generally I think the goal is to move all the modelling API definitions to StatsModels/StatsModelsBase when moving other parts of StatsBase to Statistics.

nalimilan:
@matthieugomez Out of curiosity, in what context do you use StatsModels without loading at the same time a package that depends on Distributions?

kleinschmidt:
Can you also bump the version number?  I think a patch release is fine.

matthieugomez:
@nalimilan I have also replaced the dependency of FixedEffectModels to StatsFuns instead of Distributions

matthieugomez:
Ok to merge?

PR #197, opened 2020-10-29
use trigger to run tagbot instead of cron

kleinschmidt:
To avoid disabled workflows after 60 days of inactivity, and to get instant
tagging with merged registrator PRs.  Based on [discourse
post](https://discourse.julialang.org/t/ann-required-updates-to-tagbot-yml/49249).

### DIFF
--- a/.github/workflows/TagBot.yml
+++ b/.github/workflows/TagBot.yml
@@ -1,6 +1,6 @@
 name: TagBot
 on:
-  issue_comment:  # THIS BIT IS NEW
+  issue_comment:
     types:
       - created
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/TagBot.yml
```diff
@@ -1,9 +1,11 @@
 name: TagBot
 on:
-  schedule:
-    - cron: 0 * * * *
+  issue_comment:  # THIS BIT IS NEW
```
nalimilan, line 3:
```suggestion
  issue_comment:
```

nalimilan (approved):
I knew running this every day wasn't a good idea... ;-)

.github/workflows/TagBot.yml
kleinschmidt: that's what I get for copy and pasting

PR #199, opened 2020-11-28
Switch CI to github actions and remove Travis

kleinschmidt:


palday (commented):
The failure on Future looks genuine.

While you're changing CIs, why not add

```yaml 
      - uses: julia-actions/julia-docdeploy@latest
```

to your Documentation CI? Then you can preview documentation builds. (This is what we're using in [MixedModels to preview docs](https://github.com/JuliaStats/MixedModels.jl/blob/master/.github/workflows/documenter.yml#L16).)

kleinschmidt:
I think the future failure is a change in how the small p value exponents are printing in the F/LR test tables (similar thing cropping up in GLM...):

on nightly:
```
Likelihood-ratio test: 2 models fitted on 4 observations
──────────────────────────────────────────────
     DOF  ΔDOF  Deviance  ΔDeviance  p(>Chisq)
──────────────────────────────────────────────
[1]    0         30.0000                      
[2]    1     1   10.8600   -19.1400     <1e-04
──────────────────────────────────────────────
```

tests expect:
```
Likelihood-ratio test: 2 models fitted on 4 observations
──────────────────────────────────────────────
     DOF  ΔDOF  Deviance  ΔDeviance  p(>Chisq)
──────────────────────────────────────────────
[1]    0         30.0000                      
[2]    1     1   10.8600   -19.1400      <1e-4
──────────────────────────────────────────────
```

kleinschmidt:
What happened here is that Printfageddon (JuliaLang/julia#32859) also fixed a few bugs/inconsistencies julia's `@printf`.  The `04` two-digit format for small pvalue exponents was always intended.  StatsBase.jl put the p value printing behind a version check (JuliaStats/StatsBase.jl#606).

.github/workflows/ci-future.yml
```diff
@@ -0,0 +1,47 @@
+name: CI-future
```
nalimilan, line 1: Out of curiosity, what's the advantage of using a separate workflow for nightlies?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,48 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+          - '1.0'
+          - '1' # automatically expands to the latest stable 1.x release of Julia
+        os:
+          - ubuntu-latest
+        arch:
+          - x64
```
nalimilan, line 25: I think it would make sense to test at least one 32-bit platform/version. Maybe the latest Julia release?

.github/workflows/ci-future.yml
kleinschmidt: to keep that sweet sweet green badge even when things go sideways on nighly

nalimilan: Ah, good idea. We could do that in DataFrames too, and in the PR I opened against Example.jl.

nalimilan (approved):

.github/workflows/ci-future.yml
kleinschmidt: also, less cheekily, to be able to see at a glance if any of the stable releases have failing tests AND whether nightly has failing tests, since those indicate different types of problems (only the first should block merging IMO, but we can't afford to ignore the second so we don't want to use `continue-on-error` and nighly failures green...).

kleinschmidt:
I can reproduce nightly failure on latest nightlies but it does not appear to be related to anything here...happens when trying to instantiate the project.

.github/workflows/ci-future.yml
kleinschmidt: FWIW I learned this trick from the MixedModels CI which has been on actions for a while now...

PR #200, opened 2020-12-18
fix missing methods for +

kleinschmidt:
Fills in some missing methods for adding collections of terms with `+`, and
tests which target those (formerly) missing methods.

Fixes JuliaStats/MixedModels.jl#452


palday (approved):

palday:
Looks pretty straightforward if the CIs are happy.

PR #208, opened 2021-01-19
Schema application to solitary term

### LINKED ISSUE #207: no method matching +(::InterceptTerm{true})
The following snippet runs in v0.6.15 but returns an error in v0.6.16
```julia
using StatsModels
d = (a = [1,1],)
apply_schema((ConstantTerm(1),), schema(d))
# MethodError: no method matching +(::InterceptTerm{true})
```
This change breaks the tests in [FixedEffectModels](https://github.com/FixedEffects/FixedEffectModels.jl), even though it was introduced in a minor version. Would it be possible to fix it?

palday:
Closes #207.

The fundamental problem is that `sum` always calls `+`, but just uses unary `+` when there is only one element. `reduce` just returns the only element when there is only a single element. 

Sorry about the whitespace changes -- just stripped trailing spaces in the files I edited.


kleinschmidt (approved):
LGTM; bump patch version too?

palday:
@kleinschmidt I've now also added a method for unary `+` so that `sum` works if anybody tries constructing a formula programmatically with `sum` and winds up with a singleton tuple.

PR #214, opened 2021-02-05
Fix empty tuple piracy

### LINKED ISSUE #213: StatsModels is incorrectly type pirating `show()` for empty tuples, breaking tuple display
As you can see here, `using StatsModels` causes tuples to print as an empty string:
```julia
julia> ()
()

julia> using StatsModels
[ Info: Precompiling StatsModels [3eaba693-59b7-5ba5-a881-562e759f1c8d]

julia> ()


julia> @which show(stdout, ())
show(io::IO, terms::Tuple{Vararg{Union{AbstractTerm, Tuple{Vararg{AbstractTerm,N}} where N},N}} where N) in StatsModels at /Users/nathandaly/.julia/packages/StatsModels/UtbVD/src/terms.jl:348
```

This is because empty tuples are being caught in the definition of `show()` implemented for `NTuple`s, since an empty tuple is a subtype of every NTuple:
```julia
julia> () isa NTuple{N,Int} where N
true

julia> () isa NTuple{N,String} where N
true
```

The problematic definition is here:
https://github.com/JuliaStats/StatsModels.jl/blob/5444c5e3021ec854e1464818597a463df5a4020a/src/terms.jl#L348

kleinschmidt:
Fixes #213

The type aliases for tuples of terms used `NTuple` which included empty
tuples.  This changes those aliases to use `Tuple{T, Vararg{T}}` instead, which
enforces that there's at least one `T` (`AbstractTerm` etc.) in the tuple.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "StatsModels"
 uuid = "3eaba693-59b7-5ba5-a881-562e759f1c8d"
-version = "0.6.20"
+version = "0.6.21"
 
 [deps]
 DataAPI = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
--- a/src/terms.jl
+++ b/src/terms.jl
@@ -1,6 +1,6 @@
 abstract type AbstractTerm end
-const TermOrTerms = Union{AbstractTerm, NTuple{N, AbstractTerm} where N}
-const TupleTerm = NTuple{N, TermOrTerms} where N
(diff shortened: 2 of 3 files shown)


nalimilan (approved):

NHDaly:
Awesome, thanks! 🎉 

Do note that this means that your other pirated functions, like `() + ()`, will no longer work. You might find that users were relying on this without realizing it if they had some tuples constructed dynamically and they end up empty?

Something to keep an eye on. 👍 Thanks!

kleinschmidt:
I certainly look forward to those bug reports, if only to find out what kind of strange use cases people have come up with :)

PR #218, opened 2021-02-22
Remove nonexistent export of LagTerm and dropterm

nalimilan:


kleinschmidt:
Huh, funny that these survived.  I think it's `LeadLagTerm{Union{typeof(lead), typeof(lag)}}` (maybe we should make aliases?) and `drop_term`.

nalimilan:
Given that these have been broken for a long time, aliases are probably not needed.

kleinschmidt (approved):

PR #225, opened 2021-04-25
Show R^2 and F-statistic for regression models

OndrejSlamecka:
Adds R² and F-statistic for regression models, bringing the output closer to what users might expect when coming from R.

Example:

```
julia> ols_lin
StatsModels.TableRegressionModel{LinearModel{GLM.LmResp{Vector{Float64}}, GLM.DensePredChol{Float64, LinearAlgebra.CholeskyPivoted{Float64, Matrix{Float64}}}}, Matrix{Float64}}

y ~ 1 + x

Coefficients:
──────────────────────────────────────────────────────────────────────
              Coef.  Std. Error      t  Pr(>|t|)  Lower 95%  Upper 95%
──────────────────────────────────────────────────────────────────────
(Intercept)  -442.0     54.5539  -8.10    <1e-09  -551.688   -332.312
x              51.0      1.8619  27.39    <1e-30    47.2564    54.7436
──────────────────────────────────────────────────────────────────────

R² = 0.9399
F-statistic: 750.3 on 1 and 50 degrees of freedom, p-value: 0.02898
```


src/statsmodel.jl
```diff
@@ -189,7 +189,19 @@ function Base.show(io::IO, model::TableModels)
...
+            println(io)
+            println("R²: ", round(r2(model), sigdigits=4))
+            try
+                fstat = fstatistic(model)
+                println(io, fstat)
+            catch e
+                if !(isa(e, MethodError) && e.f == fstatistic)
+                    rethrow(e)
+                end
+            end
```
palday, line 203: general speaking `if x isa Y` is a sign of non idiomatic Julia code. I would do something like 
```julia
_show_fit_stats(io::IO, model::TableModel) = nothing

function _show_fit_stats(io::IO, model::TableRegressionModel)
# stuff
end
```

and add `_show_fit_stats(io, model)` to the `show` method

src/statsmodel.jl
```diff
@@ -198,3 +210,5 @@ function Base.show(io::IO, model::TableModels)
         end
     end
 end
+
+fstatistic(m::TableRegressionModel) = fstatistic(m.model)
```
palday, line 214:
```suggestion
```

src/statsmodel.jl
```diff
@@ -122,7 +122,7 @@ const TableModels = Union{TableStatisticalModel, TableRegressionModel}
                              StatsBase.stderror, StatsBase.vcov]
 @delegate TableRegressionModel.model [StatsBase.modelmatrix,
                                       StatsBase.residuals, StatsBase.response,
-                                      StatsBase.predict, StatsBase.predict!, 
+                                      StatsBase.predict, StatsBase.predict!,
                                       StatsBase.cooksdistance]
```
palday, line 126:
```suggestion
                                      StatsBase.cooksdistance, fstatistic]
```

OndrejSlamecka:
Thanks @palday for the good suggestions, I implemented them.

src/statsmodel.jl
```diff
@@ -180,6 +180,21 @@ function StatsBase.coeftable(model::TableModels; kwargs...)
     ct
 end
 
+_show_fit_stats(io::IO, model::TableModels) = nothing
+
+function _show_fit_stats(io::IO, model::TableRegressionModel)
+    println("R²: ", round(r2(model), sigdigits=4),
+         "\t Adjusted R²: ", round(adjr2(model), sigdigits=4))
```
nalimilan, line 187: This should also be included in a `try... catch` block, right? For GLMs (or any model not implementing `r2`) this will fail.

OndrejSlamecka: I agree, changed it. Thank you.

src/statsmodel.jl
```diff
@@ -193,6 +193,22 @@ function StatsBase.coeftable(model::TableModels; kwargs...)
     ct
 end
 
+_show_fit_stats(io::IO, model::TableModels) = nothing
+
+function _show_fit_stats(io::IO, model::TableRegressionModel)
+    try
+        println("R²: ", round(r2(model), sigdigits=4),
+             "\t Adjusted R²: ", round(adjr2(model), sigdigits=4))
```
nalimilan, line 201: This should ensure that 4 digits are always used, and that "Ajusted" is aligned with the value of the F-statistic on the next line.
```suggestion
        @printf(io, "R²: %.4f  Adjusted R²: %.4f", r2(model), adjr2(model))
```

src/statsmodel.jl
```diff
@@ -193,6 +193,22 @@ function StatsBase.coeftable(model::TableModels; kwargs...)
...
+
+        fstat = fstatistic(model)
+        println(io, fstat)
```
nalimilan, line 204: `fstatistic` doesn't exist anywhere currently, and only GLM provides the needed functionality via `ftest` (https://github.com/JuliaStats/GLM.jl/pull/424/). An empty `ftest` definition should be added to StatsAPI first. Probably the API can just require that the returned object has the `dof`, `fstat` and `pval` properties so that you can extract them and print them.
```suggestion
        ftr = ftest(model)
        @printf(io, "F-statistic: %.4f on %s degrees of freedom, p-value %s",
                ftr.fstat, ftr.dof, PValue(ftr.pvalue))
```

nalimilan:
How about printing also the number of observations and degrees of freedom?

Can you also add tests?

src/statsmodel.jl
xgdgsc: @nalimilan what if adding a print API to StatsAPI so packages can implement their own printing?

palday: @xgdgsc every package is free to implement `Base.show` methods for their own types

PR #235, opened 2021-08-31
Slightly nicer error messages when a variable is missing in the data table. 

grahamstark:
A missing variable in a model can produce a very long and confusing error message if the data has many columns. This patch reproduces the nice DataFrames error missing variables error message, and produces only one of the huge messages in the stack trace rather than several (but apart from that isn't as much of an improvement as I'd intended).


kleinschmidt:
This is a great start!  It might be slightly nicer to intercept these errors somewhere other than `ModelFrame` since we're planning to phase that out.  `schema` might be a good place.  I think there are only two places where the columns are accessed during schema creation, both methods of `concrete_term`:

https://github.com/JuliaStats/StatsModels.jl/blob/50da19faf52f117fe102a1233183882c53f6f34c/src/schema.jl#L171-L174

That also seems like a reasonable place to add this check.

grahamstark:
Hi. Sorry for the delay here.  I've belatedly had a bash at adding this to  the concrete_term constructors.

src/modelframe.jl
```diff
@@ -68,9 +68,15 @@ end
 
 missing_omit(data::T, formula::AbstractTerm) where T<:ColumnTable =
     missing_omit(NamedTuple{tuple(termvars(formula)...)}(data))
-
+
```
kleinschmidt, line 71:
```suggestion

```

src/modelframe.jl
```diff
@@ -68,9 +68,15 @@ end
...
     missing_omit(NamedTuple{tuple(termvars(formula)...)}(data))
-
+    
 function ModelFrame(f::FormulaTerm, data::ColumnTable;
                     model::Type{M}=StatisticalModel, contrasts=Dict{Symbol,Any}()) where M
+    
+    msg = checknamesexist( f, data )
+    if msg != ""
+        throw(ArgumentError(msg))
+    end
```
kleinschmidt, line 78: This is now redundant with the `concrete_term` checks (since those get triggered by the `schema` call a few lines below).  Or, actually, maybe not, since `missing_omit` may be trying to access the missing columns.  Either way I don't think it really does any harm by being here, too, so you get all the missing variables at once.

src/schema.jl
```diff
@@ -166,12 +166,25 @@ julia> concrete_term(term(:a), (a = [1, 2, 3], b = [0.0, 0.5, 1.0]))
...
-    concrete_term(t, getproperty(dt, t.sym), get(hints, t.sym, nothing))
+concrete_term(t::Term, d, hints::Dict{Symbol}) = concrete_term(t, d, get(hints, t.sym, nothing))
+
+function concrete_term(t::Term, dt::ColumnTable, hint)
+    msg :: String = checkcol( dt, t.sym )
+    if msg != ""
+        throw(ArgumentError(msg))
+    end
+    return concrete_term(t, getproperty(dt, t.sym), hint)
+end
```
kleinschmidt, line 177: style: (also, we don't use type asserts like this elsewhere in the code base and the comparison below is enough to imply that it should be a string, so might just delete it)
```suggestion
function concrete_term(t::Term, dt::ColumnTable, hint)
    msg::String = checkcol( dt, t.sym )
    if msg != ""
        throw(ArgumentError(msg))
    end
    return concrete_term(t, getproperty(dt, t.sym), hint)
end
```

src/schema.jl
```diff
@@ -166,12 +166,25 @@ julia> concrete_term(term(:a), (a = [1, 2, 3], b = [0.0, 0.5, 1.0]))
...
+
+function concrete_term(t::Term, dt::ColumnTable, hints::Dict{Symbol})
+    msg :: String = checkcol( dt, t.sym )
```
kleinschmidt, line 180: same here
```suggestion
    msg::String = checkcol( dt, t.sym )
```

src/errormessages.jl
```diff
@@ -0,0 +1,44 @@
...
+`colnames` : some iterable collection of symbols
+"""
+function fuzzymatch(colnames, name::Symbol)
+        ucname = uppercase(string(name))
+        dist = [(levenshtein(uppercase(string(x)), ucname), x) for x in colnames]
+        sort!(dist)
+        c = [count(x -> x[1] <= i, dist) for i in 0:2]
+        maxd = max(0, searchsortedlast(c, 8) - 1)
+        return [s for (d, s) in dist if d <= maxd]
+end
```
kleinschmidt, line 14: indentation
```suggestion
function fuzzymatch(colnames, name::Symbol)
    ucname = uppercase(string(name))
    dist = [(levenshtein(uppercase(string(x)), ucname), x) for x in colnames]
    sort!(dist)
    c = [count(x -> x[1] <= i, dist) for i in 0:2]
    maxd = max(0, searchsortedlast(c, 8) - 1)
    return [s for (d, s) in dist if d <= maxd]
end
```

src/errormessages.jl
```diff
@@ -0,0 +1,44 @@
...
+
+"""
+Return a nice-ish error message if the Symbol `name` isn't a column name in `table`, otherwise a zero-length string.
+"""
+function checkcol(table, name :: Symbol)::String
```
kleinschmidt, line 19: Same as below...I don't think the type assert is really necessary here since it's sufficiently clear from the (short) function body that a string will be returned
```suggestion
function checkcol(table, name::Symbol)
```

src/errormessages.jl
```diff
@@ -0,0 +1,44 @@
...
+        return "There isn't a variable called '$name' in your data; the nearest names appear to be: $nearestnames"
+    end
+    return ""
+end
+
+"""
+Check that each name in the given model `f` exists in the data source `t` and return a message if not. Return a zero string otherwise.
+`t` is something that implements the `Tables` interface.
+"""
+function checknamesexist(f :: FormulaTerm, t)::String
```
kleinschmidt, line 33:
```suggestion
function checknamesexist(f::FormulaTerm, t)
```

kleinschmidt (changes requested):
Thank, this looks great once a few stylistic things are addressed!

grahamstark:
Thanks for this. I'm coming to this Mainly from Ada, hence all the types...

kleinschmidt:
Ah yeah cool!  It's a pretty common part of the julia learning curve to figure out where and how to best use type annotations.  A lot of the time they aren't even really necessary in function signatures, since the compiler is pretty good at figuring that stuff out on the fly.  The only situations when they're really necessary is when you need special handling for different types of inputs (e.g. for dispatch); otherwise the general advice is to keep the type restrictions on methods as sparse as possible.

kleinschmidt (approved):

kleinschmidt:
Thanks for the contribution!

Issue #237, opened 2021-09-11
`base` argument in `SeqDiffCoding()` does not seem to work as expected


HaiyangJin:
When `base` is set with the non-first level in the alphabetic order of the variable, it seems still to use the first level as the `base`. 

For example:
```julia
using StatsModels, DataFrames, GLM # StatsModels version: v0.6.25

df = DataFrame(IV = ["a", "b", "a", "b"],
               DV = [2, 3, 2, 5])
```

### Set `base = "a"`:
```julia
base_a = fit(LinearModel, @formula(DV ~ IV), df, 
                      contrasts=Dict(:IV => SeqDiffCoding(base = "a")))
```
```
────────────────────────────────────────────────────────────────────
             Coef.  Std. Error     t  Pr(>|t|)  Lower 95%  Upper 95%
────────────────────────────────────────────────────────────────────
(Intercept)    3.0         0.5  6.00    0.0267   0.848674    5.15133
IV: b          2.0         1.0  2.00    0.1835  -2.30265     6.30265
────────────────────────────────────────────────────────────────────
```

### Set `levels = ["a", "b"]`
```julia
level_ab = fit(LinearModel, @formula(DV ~ IV), df, 
                       contrasts=Dict(:IV => SeqDiffCoding(levels = ["a", "b"])))
```
```
────────────────────────────────────────────────────────────────────
             Coef.  Std. Error     t  Pr(>|t|)  Lower 95%  Upper 95%
────────────────────────────────────────────────────────────────────
(Intercept)    3.0         0.5  6.00    0.0267   0.848674    5.15133
IV: b          2.0         1.0  2.00    0.1835  -2.30265     6.30265
────────────────────────────────────────────────────────────────────
```
### Set `base = "b"`:
```julia
base_b = fit(LinearModel, @formula(DV ~ IV), df, 
                       contrasts=Dict(:IV => SeqDiffCoding(base = "b")))
```
```
────────────────────────────────────────────────────────────────────
             Coef.  Std. Error     t  Pr(>|t|)  Lower 95%  Upper 95%
────────────────────────────────────────────────────────────────────
(Intercept)    3.0         0.5  6.00    0.0267   0.848674    5.15133
IV: a          2.0         1.0  2.00    0.1835  -2.30265     6.30265
────────────────────────────────────────────────────────────────────
```
### Set `levels = ["b", "a"]`
```julia
level_ba = fit(LinearModel, @formula(DV ~ IV), df, 
[...]

PR #238, opened 2021-09-11
Deprecate un-used `base=` kwarg for `SeqDiffCoding`

kleinschmidt:
`SeqDiffCoding` has an un-used base kwarg and field.  This can cause confusion (#237 ), so I propose deprecating and eventually removing the kwarg.  This PR removes the internal field and adds a deprecation warning.

### DIFF
--- a/src/contrasts.jl
+++ b/src/contrasts.jl
@@ -453,7 +453,7 @@ mutable struct SeqDiffCoding <: AbstractContrasts
 end
 function SeqDiffCoding(; base=nothing, levels::Union{AbstractVector,Nothing}=nothing)
     if base !== nothing
-        Base.depwarn("`base=` kwarg for `SeqDiffCoding` is deprecated. " *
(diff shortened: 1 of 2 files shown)


src/contrasts.jl
```diff
@@ -448,7 +448,22 @@ julia> StatsModels.hypothesis_matrix(seqdiff)
...
 
-function contrasts_matrix(C::SeqDiffCoding, baseind, n)
+mutable struct SeqDiffCoding <: AbstractContrasts
+    levels::Union{AbstractVector,Nothing}
+end
+function SeqDiffCoding(; base=nothing, levels::Union{AbstractVector,Nothing}=nothing)
+    if base !== nothing
+        Base.depwarn("`base=` kwarg for `SeqDiffCoding` is deprecated. " *
+                     "Specify full order of levels using `levels=` instead",
+                     :SeqDiffCoding)
```
palday, line 458: I would be tempted to be verbose here and write "and has no effect" to really emphasize that it's pointless

palday (approved):

PR #241, opened 2021-09-21
equality and hash for terms and schemas

kleinschmidt:
based on #174 (upstream user/repo seems to have been deleted) with a few tweaks

### DIFF
--- a/src/schema.jl
+++ b/src/schema.jl
@@ -53,7 +53,7 @@ Base.merge!(a::Schema, b::Schema) = (merge!(a.schema, b.schema); a)
 Base.keys(schema::Schema) = keys(schema.schema)
 Base.haskey(schema::Schema, key) = haskey(schema.schema, key)
 
-function ==(first::Schema, second::Schema)
+function Base.:(==)(first::Schema, second::Schema)
     first === second && return true
     first.schema === second.schema && return true
     length(first.schema) != length(second.schema) && return false


src/schema.jl
```diff
@@ -53,6 +53,19 @@ Base.merge!(a::Schema, b::Schema) = (merge!(a.schema, b.schema); a)
...
 
+function Base.:(==)(first::Schema, second::Schema)
+    first === second && return true
+    first.schema === second.schema && return true
+    length(first.schema) != length(second.schema) && return false
+    for key in keys(first)
+        !haskey(second, key) && return false
+        second[key] != first[key] && return false
+    end
+    true
```
nalimilan, line 64: Isn't this identical to `first == second`? If not, maybe worth a comment. (Note that this throws if the dict contains `missing`, maybe that's OK though.)

src/schema.jl
```diff
@@ -53,6 +53,19 @@ Base.merge!(a::Schema, b::Schema) = (merge!(a.schema, b.schema); a)
 Base.keys(schema::Schema) = keys(schema.schema)
 Base.haskey(schema::Schema, key) = haskey(schema.schema, key)
 
+function Base.:(==)(first::Schema, second::Schema)
+    first === second && return true
+    first.schema === second.schema && return true
```
nalimilan, line 58: Note that this will be wrong if the dict contains `missing` (recursively). Can this happen?

src/schema.jl
```diff
@@ -53,6 +53,19 @@ Base.merge!(a::Schema, b::Schema) = (merge!(a.schema, b.schema); a)
...
+    length(first.schema) != length(second.schema) && return false
+    for key in keys(first)
+        !haskey(second, key) && return false
+        second[key] != first[key] && return false
+    end
+    true
+end
+
+Base.hash(schema::Schema, h::UInt) = hash(schema.schema, h)
```
nalimilan, line 67: Sometimes (e.g. for arrays and tuples) we add an arbitrary constant to `h` (type-specific) to ensure that `hash(schema) != hash(schema.schema)`. Not sure whether it's worth it here though.

kleinschmidt: And that can "corrupt" the type of containers because of the *ahem* zealous use of type parameters in this codebase

src/schema.jl
kleinschmidt: Hmmm I think you're right.  Probably fine to just check `first.schema == second.schema` since that's really what this check is about

src/schema.jl
kleinschmidt: As a key or a value?  Not possible either way at the moment (unless it's a pathological manuallly constructed instance)

nalimilan: As a value I think, as for keys dicts use `isequal`.

src/schema.jl
nalimilan: You don't necessarily need to hash the type: you can just define a constant and use it for all types which compare `==`.

PR #242, opened 2021-09-21
relax some checks on eltypes for contrasts and add tests

### LINKED ISSUE #236: Different String Types in Levels
With the release of CSV 0.9, the default string type being read in is `InlineString`, which leads to problems when levels for contrast codings are specified as `String`. It feels like it would be nice to allow this comparison (or possibly for subtypes of `AbstractString`).

kleinschmidt:
To maintain compat with Julia 1.0, I had to write the tests using CSV to round trip some strings...that should produce `StringN` types for versions where that's possible and do something reasonable otherwise.

fixes #236

### DIFF
--- a/test/contrasts.jl
+++ b/test/contrasts.jl
@@ -357,18 +357,13 @@
     end
 
     @testset "other string types" begin
-        using CSV
+        using WeakRefStrings
 
         using StatsModels: ContrastsMatrix
         using DataAPI: levels
 
         x = ["a", "b", "c", "a", "a", "b"]
-        io = IOBuffer()
-        tab = CSV.write(io, (; x=x, ))
-        seekstart(io)
-        x1 = Tables.columntable(CSV.File(io)).x
-        # this doesn't work w/ Julia < 1.3 because of WeakRefStrings compat
-        # x1 = WeakRefStrings.String1.(x)
+        x1 = WeakRefStrings.String1.(x)
         x1_levs = levels(x1)
 
         @test issetequal(x, x1)
--- a/test/contrasts.jl
+++ b/test/contrasts.jl
@@ -357,6 +357,8 @@
     end
 
     @testset "other string types" begin
+        using CSV
+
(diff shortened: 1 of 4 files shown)


test/contrasts.jl
```diff
@@ -355,5 +355,30 @@
...
+        # this doesn't work w/ Julia < 1.3 because of WeakRefStrings compat
+        # x1 = WeakRefStrings.String1.(x)
+        x1_levs = levels(x1)
+
+        @test issetequal(x, x1)
+
+        c1 = ContrastsMatrix(DummyCoding(), x1_levs)
+        c = ContrastsMatrix(DummyCoding(levels=["a", "b", "c"]), x1_levs)
+        @test c == c1
+        @test eltype(c.levels) == eltype(c1.levels) != eltype(c.contrasts.levels)
```
palday, line 379: For older versions, this will be equality straight across. Can you do `@static if VERSION < v"1.3"` or the like?

palday (commented):
looks pretty straightforward. Get CIs to pass and I'll be happy. This is important enough that I would also immediately tag a patch release.

kleinschmidt:
The x86 failure might be a CSV issue (https://github.com/JuliaData/CSV.jl/issues/913)

test/contrasts.jl
palday: Totally agree.

nalimilan (approved):

nalimilan:
In theory  think the rule is to bump the minor version when dropping support for older Julia versions, in case one needs to do another patch release to fix a bug there too. But making a breaking release that requires updating version bounds for all dependencies would be annoying.

ararslan:
Seems unlikely that we'd need to backport things to make another release that supports unsupported Julia versions, as 1.6 is LTS.

kleinschmidt:
I'd like to merge the fix in #252 first, then merge that here, make sure CI still passes, adn then we can merge.

Issue #220, opened 2021-04-04
`fit` is very slow for new formulas


evanfields:
Calling `StatsModels.fit` with a not yet seen formula seems to trigger pretty slow compilation, even if a structurally equivalent formula with different names _has_ been seen before. Triggering `fit` with a formula which has been seen before is very fast.

The below reproducing example using `GLM` and `DataFrames`, and closely mimics how I stumbled upon this issue in the wild. I'm not familiar with the StatsModels/GLM internals, but if this example isn't minimal enough I can try to drill down.
```julia
julia> using StatsModels, DataFrames, GLM

julia> df = DataFrame(rand(100,100));

julia> function f(df, a, b, c)
           reg_form = Term(a) ~ Term(b) + Term(c)
           return r2(lm(reg_form, df))
       end
f (generic function with 1 method)

julia> # precompile f with arg types dataframe, symbol, symbol, symbol

julia> f(df, :x1, :x2, :x3)
0.0033857274581728936

julia> # That was really slow, around 7 secons. Future calls with the exact same symbols are fast

julia> # call f with the same arg types, but new symbol values

julia> @elapsed f(df, :x1, :x2, :x4)
0.1441765

julia> # call f again with already-seen symbol values, and it's super fast

julia> @elapsed f(df, :x1, :x2, :x4)
0.00016
       
julia> versioninfo()
Julia Version 1.6.0
Commit f9720dc2eb (2021-03-24 12:55 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: AMD Ryzen 5 1600X Six-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, znver1)
Environment:
  JULIAPATH = C:\Users\ejfie\AppData\Local\Programs\Julia-1.6.0\bin
  JULIA_NUM_THREADS = 4

(@v1.6) pkg> st StatsModels
      Status `C:\Users\ejfie\.julia\environments\v1.6\Project.toml`
  [3eaba693] StatsModels v0.6.21
  
julia> using Profile

julia> @profile begin
           for i in 5:20
               symb = Symbol("x$i")
               f(df, :x1, :x2, symb)
           end
       end

julia> Profile.print(mincount=50,maxdepth=30)
Overhead ╎ [+additional indent] Count File:Line; Function
=========================================================
   ╎1254 @Base\client.jl:485; _start()
[...]

PR #247, opened 2021-10-15
WIP use Tables.Columns instead of `columntable`

kleinschmidt:
This uses `Tables.Columns` as a (potentially) lightweight wrapper around input tables that does not convert them to the strongly typed `NamedTuple` of `Vector`s representation.  This might make some things easier on the compiler (e.g. #220 ).

Requires Tables 1.6.0 since that's when `Columns` stopped being a lie ;)

There's some design issues to work out here still, since a generic `NamedTuple` could be EITHER a column table (if it contains vectors) or a single row, and there are a handful of methods that specialize on that to provide special handling (most notably `modelcols(::InteractionTerm, ...)`).  What we PROBABLY will need to do is to add parallel methods for `Row` in a similar fashion, but I'm not sure about that.  In the mean time, merging this would be breaking since you lose first class support for named tuples of singletons, which is part of the current public API.  There may be a way around that but I haven't dug into it yet...

### DIFF
--- a/src/terms.jl
+++ b/src/terms.jl
@@ -488,12 +488,12 @@ modelcols(t::Term, d) =
                         "Use apply_schema to create concrete terms first"))
 
 # TODO: @generated to unroll the getfield stuff
-modelcols(ft::FunctionTerm{Fo,Fa,Names}, d::Columns) where {Fo,Fa,Names} =
+modelcols(ft::FunctionTerm{Fo,Fa,Names}, d) where {Fo,Fa,Names} =
     ft.fanon.(getcolumn.(Ref(d), Names)...)
 
-modelcols(t::ContinuousTerm, d::Columns) = copy.(getcolumn(d, t.sym))
+modelcols(t::ContinuousTerm, d) = copy.(getcolumn(d, t.sym))
 
-modelcols(t::CategoricalTerm, d::Columns) = t.contrasts[getcolumn(d, t.sym), :]
+modelcols(t::CategoricalTerm, d) = t.contrasts[getcolumn(d, t.sym), :]
 
 
 """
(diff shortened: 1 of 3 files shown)


src/schema.jl
```diff
@@ -112,16 +112,16 @@ julia> sch[term(:y)]
...
+    schema(Term.(collect(Tables.columnnames(data))), data, hints)
 
 # handle hints:
-schema(ts::AbstractVector{<:AbstractTerm}, dt::ColumnTable,
-      hints::Dict{Symbol}=Dict{Symbol,Any}()) =
-    sch = Schema(t=>concrete_term(t, dt, hints) for t in ts)
+function schema(ts::AbstractVector{<:AbstractTerm},
+                data,
+                hints::Dict{Symbol}=Dict{Symbol,Any}())
+    data = Tables.Columns(Tables.columns(data))
```
nalimilan, line 122: What is the advantage of wrapping the result of `Tables.columns` in a `Tables.Columns` object?

src/schema.jl
```diff
@@ -168,15 +168,15 @@ a(continuous)
...
+concrete_term(t::Term, d, hint) =
+    concrete_term(t, getcolumn(d, t.sym), hint)
+concrete_term(t::Term, d, hints::Dict{Symbol}) =
+    concrete_term(t, getcolumn(d, t.sym), get(hints, t.sym, nothing))
 concrete_term(t::Term, d) = concrete_term(t, d, nothing)
 
 # if the "hint" is already an AbstractTerm, use that
 # need this specified to avoid ambiguity
-concrete_term(t::Term, d::ColumnTable, hint::AbstractTerm) = hint
+concrete_term(t::Term, d::Tables.Columns, hint::AbstractTerm) = hint
```
nalimilan, line 179: Why not just
```suggestion
concrete_term(t::Term, d, hint::AbstractTerm) = hint
```

nalimilan:
> There's some design issues to work out here still, since a generic `NamedTuple` could be EITHER a column table (if it contains vectors) or a single row, and there are a handful of methods that specialize on that to provide special handling (most notably `modelcols(::InteractionTerm, ...)`). What we PROBABLY will need to do is to add parallel methods for `Row` in a similar fashion, but I'm not sure about that. In the mean time, merging this would be breaking since you lose first class support for named tuples of singletons, which is part of the current public API. There may be a way around that but I haven't dug into it yet...

For which situations do we need `modelcols` to take a row? For clarity, we could require row objects to be `Tables.AbstractRow` even if Tables.jl doesn't require that. Otherwise confusing things could happen (weird errors when the `Tables.istable` check fails for some reason...).

PR #252, opened 2022-02-28
Fix missing `coeftable` method detection.

palday:
I think this is the result of some of the re-organization going on with StatsBase/StatsAPI.

Bottom line is that `coeftable` used to have a default method with `error`, but this is now missing and so we get a `MethodError`.

### DIFF
--- a/src/statsmodel.jl
+++ b/src/statsmodel.jl
@@ -132,7 +132,7 @@ const TableModels = Union{TableStatisticalModel, TableRegressionModel}
                              StatsBase.stderror, StatsBase.vcov, StatsBase.fitted]
 @delegate TableRegressionModel.model [StatsBase.modelmatrix,
                                       StatsBase.residuals, StatsBase.response,
-                                      StatsBase.predict, StatsBase.predict!, 
+                                      StatsBase.predict, StatsBase.predict!,
                                       StatsBase.cooksdistance]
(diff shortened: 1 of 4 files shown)


src/statsmodel.jl
```diff
@@ -195,16 +195,15 @@ end
...
-        println(io)
-        println(io, model.mf.f)
-        println(io)
         println(io,"Coefficients:")
-        show(io, ct)
+        show(io, coeftable(model))
     catch e
-        if isa(e, ErrorException) && occursin("coeftable is not defined", e.msg)
+        if isa(e, MethodError) || isa(e, ErrorException) && occursin("coeftable is not defined", e.msg)
             show(io, model.model)
```
palday, line 207: In addition to checking for a `MethodError` (so that we support the last several StatsBase/StatsAPI versions), I re-arranged things to that only the `coeftable` was in the try/catch block. Then we skip the re-throw for errors happening in the other statements.

kleinschmidt (approved):
looks good if CI is happy. since this is a bugfix I think we might also want to bump the version.

palday:
I'll do the version bump -- CI fail on nightly is definitely a Julia and not StatsModels bug (double free!)

palday:
Julia 1.0 failure is something in SpecialFunctions. I'm pro ignoring that failure b/c it's not StatsModels and our Julia 1.0 support is going away very soon.

Issue #250, opened 2021-11-10
lrtest fails on doc string example / for GLM Models


awadell1:
Per the doc strings for `lrtest`:

```julia
julia> using DataFrames, GLM
  
julia> dat = DataFrame(Result=[1, 0, 1, 1, 1, 0, 0, 0, 1, 0, 1, 1],
                       Treatment=[1, 1, 1, 2, 2, 2, 1, 1, 1, 2, 2, 2],
                       Other=categorical([1, 1, 2, 1, 2, 1, 3, 1, 1, 2, 2, 1]));

julia> nullmodel = glm(@formula(Result ~ 1), dat, Binomial(), LogitLink());

julia> model = glm(@formula(Result ~ 1 + Treatment), dat, Binomial(), LogitLink());

julia> bigmodel = glm(@formula(Result ~ 1 + Treatment + Other), dat, Binomial(), LogitLink());

julia> lrtest(nullmodel, model, bigmodel)
Likelihood-ratio test: 3 models fitted on 12 observations
──────────────────────────────────────────────
     DOF  ΔDOF  Deviance  ΔDeviance  p(>Chisq)
──────────────────────────────────────────────
[1]    1         16.3006
[2]    2     1   15.9559    -0.3447     0.5571
[3]    4     2   14.0571    -1.8988     0.3870
──────────────────────────────────────────────

julia> lrtest(bigmodel, model, nullmodel)
Likelihood-ratio test: 3 models fitted on 12 observations
──────────────────────────────────────────────
     DOF  ΔDOF  Deviance  ΔDeviance  p(>Chisq)
──────────────────────────────────────────────
[1]    4         14.0571
[2]    2    -2   15.9559     1.8988     0.3870
[3]    1    -1   16.3006     0.3447     0.5571
──────────────────────────────────────────────
```

However, `categorical` is not defined (Shows up in doctests too: https://github.com/JuliaStats/StatsModels.jl/runs/4027286993?check_suite_focus=true#step:4:2433). Adding a `using CategoricalArrays` also errors with:

```julia
julia> using DataFrames, GLM, CategoricalArrays

julia> dat = DataFrame(Result=[1, 0, 1, 1, 1, 0, 0, 0, 1, 0, 1, 1],
                                     Treatment=[1, 1, 1, 2, 2, 2, 1, 1, 1, 2, 2, 2],
                                     Other=categorical([1, 1, 2, 1, 2, 1, 3, 1, 1, 2, 2, 1]));

[...]

PR #259, opened 2022-07-30
fix docstring for lrtest

jerlich:
changing `categorical` to `string.` fixes issue #250


nalimilan (approved):

nalimilan:
Thanks! It would be a good idea to make doctest failures into errors so that this kind of thing doesn't go unnoticed in the future.

Issue #260, opened 2022-07-31
lrtest incorrect


jerlich:
I initially posted this issue (incorrectly) to `GLM.jl`
https://github.com/JuliaStats/GLM.jl/issues/490

`lrtest` tests a difference in deviance, but the correct definition for the test statistic is twice the difference in the loglikelihood. 

I have edited lrtest to be consistent with `lmtest::lrtest` in R. But this breaks tests.
in lrtest.jl lines 108:119 become
```
    dev = deviance.(mods)
    Δdev = _diff(dev)

    loglik = loglikelihood.(mods)
    Δloglik = _diff(loglik)

    Δdf = _diff(df)
    dfr = Int.(dof_residual.(mods))

    if (forward && any(x -> x > 0, Δdev)) || (!forward && any(x -> x < 0, Δdev))
        throw(ArgumentError("Residual deviance must be strictly lower " *
                            "in models with more degrees of freedom"))
    end

    pval = (NaN, chisqccdf.(abs.(Δdf), 2 .* abs.(Δloglik))...)
    
```
I'm curious how the 'correct' pvals for tests were set.

ararslan:
Tangentially relevant but "residual deviance must be strictly lower in models with more degrees of freedom" seems not to be the case in practice for all types of models, and is (at least partially?) why MixedModels.jl defines its own likelihood ratio test function.

nalimilan:
> I'm curious how the 'correct' pvals for tests were set.

The test in StatsModels is really just a basic check that code works. We should put more thorough tests in GLM to cover all model families and cross check against other implementations.

palday:
Why MixedModels.jl defines its own LRT:
1. We needed it while the `lrtest` PR was stalled and so we made it and released it.
[...]

PR #261, opened 2022-08-02
Fix `lrtest` for model families with dispersion

nalimilan:
`lrtest` relied on the deviance rather than the log-likelihood, which is not correct for model families where a dispersion parameter needs to be taken into account. Scaling the deviance would be more efficient than computing the log-likelihood via `loglikelihood`, but there is currently no generic API for this and this may not work for non-GLM models.
We could imagine defining a `likelihoodratio(m1, m2) = loglikelihood(m1) - loglikelihood(m2)` method that packages could override for performance, but this may not be worth it.

Also make the check that more complex models have a strictly better fit than simpler nested ones less strict. The more complex model may have the same deviance, and due to approximations it may even have a slightly higher deviance.

Fixes https://github.com/JuliaStats/GLM.jl/issues/490, https://github.com/JuliaStats/StatsModels.jl/issues/260.

### DIFF
--- a/src/lrtest.jl
+++ b/src/lrtest.jl
@@ -1,6 +1,7 @@
 struct LRTestResult{N}
     nobs::Int
     deviance::NTuple{N, Float64}
+    loglikelihood::NTuple{N, Float64}
     dof::NTuple{N, Int}
     pval::NTuple{N, Float64}
 end
@@ -24,8 +25,8 @@ test to determine if the first one fits significantly better than the next.
 
 A table is returned containing degrees of freedom (DOF),
 difference in DOF from the preceding model, deviance, difference in deviance
-from the preceding model, and likelihood ratio and p-value for the comparison
-between the two models.
+from the preceding model, and chi-squared statistic (i.e. absolute value of
+twice the log of likelihood ratio) and p-value for the comparison between the two models.
 
 Optional keyword argument `atol` controls the numerical tolerance when testing whether
 the models are nested.
@@ -111,37 +112,48 @@ function lrtest(mods::StatisticalModel...; atol::Real=0.0)
     Δdf = _diff(df)
     dfr = Int.(dof_residual.(mods))
 
-    if (forward && any(x -> x > 0, Δdev)) || (!forward && any(x -> x < 0, Δdev))
-        throw(ArgumentError("Residual deviance must be strictly lower " *
-                            "in models with more degrees of freedom"))
+    ll = loglikelihood.(mods)
+    chisq = (NaN, 2 .* abs.(_diff(ll))...)
+
+    for i in 2:length(dev)
+        if ((forward && dev[i-1] < dev[i]) ||
+            (!forward && dev[i-1] > dev[i])) &&
+            dev[i-1] ≉ dev[i]
+               throw(ArgumentError("Residual deviance must not be larger " *
+                                   "in models with more degrees of freedom"))
+        end
     end
 
-    pval = (NaN, chisqccdf.(abs.(Δdf), abs.(Δdev))...)
-    return LRTestResult(Int(nobs(mods[1])), dev, df, pval)
+    pval = chisqccdf.(abs.(Δdf), chisq)
+    return LRTestResult(Int(nobs(mods[1])), dev, ll, df, pval)
 end
 
 function Base.show(io::IO, lrr::LRTestResult{N}) where N
     Δdf = _diff(lrr.dof)
     Δdev = _diff(lrr.deviance)
+    chisq = abs.(_diff(lrr.loglikelihood))
 
-    nc = 6
+    nc = 7
     nr = N
     outrows = Matrix{String}(undef, nr+1, nc)
 
-    outrows[1, :] = ["", "DOF", "ΔDOF", "Deviance", "ΔDeviance", "p(>Chisq)"]
+    outrows[1, :] = ["", "DOF", "ΔDOF", "Deviance", "ΔDeviance", "Chisq", "p(>Chisq)"]
 
     outrows[2, :] = ["[1]", @sprintf("%.0d", lrr.dof[1]), " ",
-                     @sprintf("%.4f", lrr.deviance[1]), " ", " "]
+                     @sprintf("%.4f", lrr.deviance[1]),
+                     " ", " ", " "]
 
     for i in 2:nr
         outrows[i+1, :] = ["[$i]", @sprintf("%.0d", lrr.dof[i]),
                            @sprintf("%.0d", Δdf[i-1]),
-                           @sprintf("%.4f", lrr.deviance[i]), @sprintf("%.4f", Δdev[i-1]),
-                           string(StatsBase.PValue(lrr.pval[i])) ]
+                           @sprintf("%.4f", lrr.deviance[i]),
+                           @sprintf("%.4f", Δdev[i-1]),
+                           @sprintf("%.4f", chisq[i-1]),
+                           string(StatsBase.PValue(lrr.pval[i]))]
     end
     colwidths = length.(outrows)
     max_colwidths = [maximum(view(colwidths, :, i)) for i in 1:nc]
-    totwidth = sum(max_colwidths) + 2*5
+    totwidth = sum(max_colwidths) + 2*6
 
     println(io, "Likelihood-ratio test: $N models fitted on $(lrr.nobs) observations")
     println(io, '─'^totwidth)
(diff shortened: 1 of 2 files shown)


src/lrtest.jl
```diff
@@ -111,37 +112,48 @@ function lrtest(mods::StatisticalModel...; atol::Real=0.0)
...
-    if (forward && any(x -> x > 0, Δdev)) || (!forward && any(x -> x < 0, Δdev))
-        throw(ArgumentError("Residual deviance must be strictly lower " *
-                            "in models with more degrees of freedom"))
+    ll = loglikelihood.(mods)
+    chisq = (NaN, 2 .* abs.(_diff(ll))...)
+
+    for i in 2:length(dev)
+        if ((forward && dev[i-1] < dev[i]) ||
+            (!forward && dev[i-1] > dev[i])) &&
+            dev[i-1] ≉ dev[i]
```
ararslan, line 121: Should the user be able to pass a tolerance value for the approximate equality here?

Also, I don't yet fully understand why (and it could be that I've done something wrong), but this still seems not to be true for beta regression. As an example, using the data defined [here](https://github.com/ararslan/BetaRegression.jl/blob/main/test/runtests.jl#L42-L80), I get
```julia
julia> model = fit(BetaRegressionModel, @formula((food / income) ~ 1 + income + persons), data);

julia> nullmodel = fit(BetaRegressionModel, @formula((food / income) ~ 1), data);

julia> (deviance(model), dof(model))
(37.18054389739939, 4)

julia> (deviance(nullmodel), dof(nullmodel))
(36.36829121985548, 2)
```
Compare that to R:
```r
> library(betareg)
> data(FoodExpenditure, package="betareg")
> model <- betareg(I(food / income) ~ 1 + income + persons, data=FoodExpenditure, link="logit", link.phi="identity")
> nullmodel <- betareg(I(food / income) ~ 1, data=FoodExpenditure, link="logit", link.phi="identity")
> sum(residuals(model, type="deviance")^2)
37.18054
> nobs(model) - df.residual(model)
4
> sum(residuals(nullmodel, type="deviance")^2)
36.36829
> nobs(nullmodel) - df.residual(nullmodel)
2
```

src/lrtest.jl
```diff
@@ -24,8 +25,8 @@ test to determine if the first one fits significantly better than the next.
 
 A table is returned containing degrees of freedom (DOF),
 difference in DOF from the preceding model, deviance, difference in deviance
-from the preceding model, and likelihood ratio and p-value for the comparison
-between the two models.
+from the preceding model, and chi-squared statistic (i.e. absolute value of
+twice the log of likelihood ratio) and p-value for the comparison between the two models.
```
palday, line 29:
```suggestion
twice the log of the likelihood ratio) and p-value for the comparison between the two models.
```

or 

```suggestion
twice the difference in log likelihood) and p-value for the comparison between the two models.
```

src/lrtest.jl
```diff
@@ -24,8 +25,8 @@ test to determine if the first one fits significantly better than the next.
 
 A table is returned containing degrees of freedom (DOF),
 difference in DOF from the preceding model, deviance, difference in deviance
-from the preceding model, and likelihood ratio and p-value for the comparison
-between the two models.
+from the preceding model, and chi-squared statistic (i.e. absolute value of
```
palday, line 28:
```suggestion
from the preceding model, chi-squared statistic (i.e. absolute value of
```

palday (approved):

src/lrtest.jl
palday: Quoting [myself](https://github.com/JuliaStats/StatsModels.jl/issues/260#issuecomment-1202888917):

> Part of the fundamental difficulty with mixed models -- both computationally and in terms of users' understanding -- is the way foundational regression theory is taught. Many equalities that hold for a particular subset of the exponential family are taught as the actual definitions of things like deviance, residual degrees of freedom, etc., instead of special cases. (@ararslan you've also felt this with beta regression!)
> 
> Addressing @ararslan's comment: I don't remember the default `isnested` definition, but maybe the "residual deviance must be strictly lower in models with more degrees of freedom" check should be part of `isnested` instead of `lrtest` itself. Or maybe some other function that can have methods defined for new model types without defining needing to specialize `lrtest`.

src/lrtest.jl
jerlich: I think the 2nd is clearer.

src/lrtest.jl
nalimilan: `isnested` has no fallback definition. I think I added that check on the deviance being smaller as a sanity check for models that do not define `isnested` (yet). I could disable it when `isnested` is defined, trusting implementations to enforce it. But results would become meaningless if that check doesn't hold (at least with some tolerance), so keeping it would make sense. It could be based on the log-likelihood rather than the deviance though, so that its relation to the test statistic is more direct.

Regarding beta regression, I'm not sure why the null model has a lower deviance. Maybe that's due to the fact that they have a different dispersion, so that deviance should be scaled accordingly? At least it has a lower log-likelihood, which is reassuring.

Maybe we should stop reporting the deviance altogether when printing `LRTestResult` objects, and instead show the likelihood? R's `anova(..., test="LRT")` prints the deviance, but it uses the scaled deviance to perform the test, which can be misleading (the scaling factor isn't printed...). We could keep the internal field for backward compatibility and only change the printing.

jerlich: Since the test statistic is using loglikelihood, it makes more sense (to me) to report it in the output (instead of deviance). MATLAB has a quite verbose output when comparing models. 

```

    Theoretical Likelihood Ratio Test

    Model    DF    AIC       BIC       LogLik     LRStat     deltaDF    pValue 
    lm1      4     44.602    55.023    -18.301                                 
    lm2      5     45.952    58.978    -17.976    0.65029    1          0.42001
```

nalimilan: I've also checked Stata's `lrtest` and R's `lmtest::lrtest`, they also report similar statistics as those reported by MATLAB. So we can probably drop the deviance. I'm not sure about the AIC and the BIC (there's also the AICc...).

nalimilan: Let's print both the log-likelihood and the deviance for now to limit disruption for users. I've updated the check that more complex models have a lower deviance to use the log-likelihood instead.

nalimilan:
@palday I've noticed that MixedModels.jl uses the deviance for `likelihoodratiotest`, except for linear models where `-2loglikelihood` is used. AFAICT this doesn't give correct results for GLMs with a dispersion parameter, such as Gamma, like for `lrtest`, right? Do you also plan to switch to using `loglikelihood`?

src/lrtest.jl
ararslan: Updating to use log likelihood instead of deviance means BetaRegression now Just Works™ with `lrtest` and produces results identical to R's `lrtest`. 🥳 Thanks Milan!

src/lrtest.jl
```diff
@@ -106,42 +107,53 @@ function lrtest(mods::StatisticalModel...; atol::Real=0.0)
...
-    if (forward && any(x -> x > 0, Δdev)) || (!forward && any(x -> x < 0, Δdev))
-        throw(ArgumentError("Residual deviance must be strictly lower " *
-                            "in models with more degrees of freedom"))
+    ll = loglikelihood.(mods)
+    chisq = (NaN, 2 .* abs.(_diff(ll))...)
+
+    for i in 2:length(ll)
+        if ((forward && ll[i-1] > ll[i]) ||
+            (!forward && ll[i-1] < ll[i])) &&
+            ll[i-1] ≉ ll[i]
```
ararslan, line 120: This was actually part of my earlier comment but got lost because the remainder of the comment was beefy. Should we allow the user to provide a tolerance for this check? We already have an `atol` argument used for checking whether the models are nested, perhaps it would make sense to reuse that here?

palday:
@nalimilan we should change (and will :wink:), although currently we don't support GLMMs with a dispersion parameter:

https://github.com/JuliaStats/MixedModels.jl/blob/ea9b86eb816f4047e59aaa1a3d78a390c268ee3f/src/generalizedlinearmixedmodel.jl#L373-L377

and so we don't have to worry about the comparison to the GLM deviance. (The problem here with deviance vs. loglikelihood is actually deeply intertwined with the problem of fitting GLMMs with a dispersion parameter.)

For LMM, it doesn't matter since the "deviance" is indeed just the objective, which is -2 loglikelihood. (In practice, we can actually compute -2 loglikelihood directly and then use that to compute the loglikelihood.)

src/lrtest.jl
nalimilan: Yeah why not. I've pushed a commit to do that. BTW I discovered that the condition was backwards!

nalimilan:
Can you make a release if you're OK with the PR? I won't be on my computer for the next two weeks.

ararslan (approved):

palday (approved):

PR #266, opened 2022-10-24
Reduce allocations

rafaqz:
This PR reduces array allocations for some nice performance improvements, in two main places.

1. It makes `missing_omit` `_maybe_missing_omit`, which does nothing when none of the columns have `Missing` eltypes.
2. It uses lazy generators in term.jl instead of array generators.


src/modelframe.jl
```diff
@@ -53,22 +53,26 @@ end
 _missing_omit(x::AbstractVector{T}) where T = copyto!(similar(x, nonmissingtype(T)), x)
```
ararslan, line 53: If I recall, the fact that this makes a copy even if `T` doesn't allow `Missing` was intentional to ensure that every column gets copied and downstream users of this function can mutate the result without worrying about affecting the original data. If we don't care about that, we can do
```suggestion
_missing_omit(x::AbstractVector{T}) where {T>:Missing} = copyto!(similar(x, nonmissingtype(T)), x)
_missing_omit(x::AbstractVector{T}) where {T} = x
```
without needing any other changes.

PR #272, opened 2023-01-24
Update doctests and make CI fail if they don't pass

nalimilan:
We cannot just use `strict=true` currently as some docstrings are considered as missing.


kleinschmidt (commented):
doctest updates look good, but I'm not seeing the "make CI fail" part...

nalimilan:
Woops. Should be fixed now.

kleinschmidt (commented):
I sorta thought that we _didn't_ want to fail our builds for failing doctests, since they're not _just_ testing stuff in our package but printing in base/other packages which can change without a breaking version bump.

At some point I think we'd discussed using something liek reviewdog + `doctest(; fix=true)` to surface changes during PR review.  That maybe would catch some churn from the filtered things that we know will change (dict ordering etc.) but would make it possible to keep the doctests up-to-date without suprious build failures unrelated to PR changes...

nalimilan:
Note that doctest failures are already annoying in PRs as they are shown in the files view, even for untouched files.

kleinschmidt:
We could add this action as a workflow step:

https://github.com/julia-actions/julia-fix-doctests

it works on PRs where there's a "fix doctests" label (and removes teh label afterwards...)

kleinschmidt:
merged master and will re-run CI

kleinschmidt (approved):

PR #280, opened 2023-03-13
update NEWS.md for #183

kleinschmidt:
I think I got the main things from #183.  Main thing here is to call out run-time function terms and the "standard" upgrade path for package developers using special syntax.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -27,6 +27,11 @@
     non-special call).  The above recommendation may still Just Work™ but some
     thorough testing is recommended.
 
+  - It is not possible (although not exactly _convenient_) to construct
+    `FunctionTerm`s at run-time (i.e., outside of a `@formula` macro).  See [the
+    tests](https://github.com/JuliaStats/StatsModels.jl/blob/623906fa27ce84a1a2a5e62014d6b9f58d2ccd47/test/protect.jl#L23-L24)
+    for an example of how this might be accomplished in practice.
+
   - Special syntax is introduced to `protect` function calls from the usual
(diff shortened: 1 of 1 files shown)


NEWS.md
```diff
@@ -1,18 +1,62 @@
...
+    ```
+    
+    However, if the `@formula` syntax (i.e., `&` for interaction, `+` for union,
+    etc.) should apply _within_ the call to `my_fun`, then care needs to be
+    taken to un-protect any nested calls as necessary (although this is only
+    likely to be a concern if the special syntax occurs as a _child_ of another,
+    non-special call).  The above recommendation may still Just Work™ but some
+    thorough testing is recommended.
+
+  - It is not possible (although not exactly _convenient_) to construct
```
palday, line 30:
```suggestion
  - It is now possible (although not exactly _convenient_) to construct
```

NEWS.md
```diff
@@ -1,18 +1,62 @@
...
+    `@formula` specials (i.e., to take the logarithm of an interaction term, use
+    `log(unprotect(a & b))`).
+
 - Various minor but breaking changes in contrast coding: (#273)
+
   - `ContrastsMatrix` can use arbitrary `AbstractMatrix`es to store the actual
     contrasts matrix.
+
   - The number and order of type parameters on `ContrastsMatrix{C,M,T,U}` have
     changed (from `ContrastsMatrix{C,T,U}`, with the addition of the second type
```
palday, line 47:
```suggestion
    changed (from `ContrastsMatrix{C,T,U}`), with the addition of the second type
```

palday (approved):

PR #288, opened 2023-04-04
use released GLM in docs now

kleinschmidt:
Ultimately we want to remove GLM-based things from the docs since it means we can't build docs with GLM-breaking changes, but for now this will allow docs to build.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -15,7 +15,6 @@ jobs:
       - uses: julia-actions/setup-julia@v1
         with:
           version: '1.8'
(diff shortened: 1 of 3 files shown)


palday (approved):

palday:
When GLM 2.0 comes out and drops its use of `TableRegressionModel`, then we can more clearly distinguish which package provides which functionality and focus on the stuff StatsModels actually vendors.

kleinschmidt:
Ugh, the doctests that were updated in #272 were based on https://github.com/JuliaStats/GLM.jl/pull/339 I think but that is not compatible with StatsModels 0.7 so I've had to revert all those doctest changes to the uglier ones.

PR #291, opened 2023-05-02
bump StatsBase compat

palday:


palday:
@kleinschmidt should we backport this to the 0.6 series?

ararslan (approved):

kleinschmidt:
> @kleinschmidt should we backport this to the 0.6 series?

Hmmmmmm I don't really think it's worth it unless someone complains.

Issue #293, opened 2023-05-26
modelcols converts x matrix of Float32 to Float64


xgdgsc:
I' m attempting https://github.com/JuliaStats/GLM.jl/issues/260 . After some line running of code, it seems modelcols converts x matrix of Float32 to Float64, while keeping y Float32, which might be the cause. Any suggestions on fixing this behavior?

kleinschmidt:
Without a reproducer it's a bit hard to tell _exactly_ what you're trying to do, but here's my understanding of the current `modelcols` situation:

- `ContinuousTerm`s are just copied, which should not on its own change eltype (this why a continuous left-hand side of a formula is un-changed, I think)
- `CategoricalTerm`s look up rows in the associated contrasts matrix.  As of #273 , the contrasts matrix can be any AbstractMatrix, so if you have categorical variables you can provide your own contrasts matrix; I'm not sure off teh top of my head whether/how you'd actually thread through a non-`Matrix{Float64}` contrasts matrix but it should be possible, using `StatsModels.ConstrastsCoding` as a fallback.
- `InteractionTerm`, ah now here is one issue: this creates either `ones(nrows)` (for `InterceptTerm{true}`, or an empty `Matrix{Float64}` (for `InterceptTerm{false}`).  and that, combined with..
- `MatrixTerm` (which is what the collection of terms on the RHS of the formula becomes) does `hcat(modelcols.(terms)...)` essentially, which is gonna trigger julia's normal promotion rules and convert everything to Float64.

So one pretty straightforward fix would be to make the columns generated by `InterceptTerm` be some "weak" type (like `Bool`) that will never cause promotion of other columns.  That may be enough to make this all Just Work™...

kleinschmidt:
Long term, I think we'd probably want to support a third `T` argument to `modelcols` that converts things as needed, but that's a bit trickier to get right (have to thread through lots of different code paths, including external libraries that extend the syntax...)

kleinschmidt:
Another possible fix would be to introduce `modelcols!(storage, term, data)` methods which would allow uesrs to pass in their own typed containers (using `StatsModels.width` to figure out the appropriate size).

xgdgsc:
Thanks. I tested the branch dfk/float32s with 
```
using GLM
using Random
using DataFrames
x32 = randn(Float32, (100, 1))
y32 = randn(Float32, 100)
regdata = DataFrame(:x => x32[:, 1], :y => y32)
[...]

Issue #271, opened 2023-01-24
roadmap to 1.0 release


kleinschmidt:
Current plan is that we'll tag 0.7 ASAP (today probably) and then we commit to tagging 1.0 in one month (so Feb 24).  #183 is sufficiently breaking taht we want to make sure that dependencies have time to upgrade and shake out any sneaky bugs that may require further breakage before we commit to 1.0.

For 0.7:
- [ ] #183 

For 1.0: (with possible 0.8 if we want to give dependencies a release to target for upgrades) 
- [ ] remove table regression model
- [ ] maybe #247 
- [ ] #175

palday:
Packages we need to check (feel free to edit/add to)

- [ ] MixedModels.jl https://github.com/JuliaStats/MixedModels.jl/pull/664
- [ ] Effects.jl
- [ ] GLM.jl
- [ ] RegressionFormulae.jl
- [ ] FixedEffectsModels.jl

PharmCat:
Hi! Can you kindly clarify - can  #154 be included in 1.0 release plan?  I suppose `apply_context` stage can close some other issues as  #268 ...

PharmCat:
Here one PR I make for GLM.jl : https://github.com/JuliaStats/GLM.jl/pull/508 for Testing of Between-Subjects Effects, there one problem:  it relies very heavily on the internal structures of the types from StatsModels. But I didn't found in StatsModels  appropriate  functions with methods to access these quantities in a stable way for this code:

```
"""
    lcontrast(obj, i::Int)

L-contrast matrix for `i` fixed effect.
"""
function lcontrast(obj, i::Int)
    n = length(obj.mf.f.rhs.terms)
    if i > n || n < 1 error("Factor number out of range 1-$(n)") end
    p    = size(obj.mm.m, 2) # number of coefs
    inds = findall(x -> x==i, obj.mm.assign)
    if typeof(obj.mf.f.rhs.terms[i]) <: CategoricalTerm
        mxc   = zeros(size(obj.mf.f.rhs.terms[i].contrasts.matrix, 1), p)
        mxcv  = view(mxc, :, inds)
        mxcv .= obj.mf.f.rhs.terms[i].contrasts.matrix
        mx    = zeros(size(obj.mf.f.rhs.terms[i].contrasts.matrix, 1) - 1, p)
        for i = 2:size(obj.mf.f.rhs.terms[i].contrasts.matrix, 1) # correct for zero-intercept model
[...]

PR #294, opened 2023-05-26
make `InterceptTerm`s generate `Bool`s to avoid unnecessary promotion of other columns

kleinschmidt:
This is one possible fix for #293.  As described in that thread, the bottleneck here is actually that `InterceptTerm`s generate `Array{Float64}`s, which then get `hcat` with other columns and promote them to Float64.  With this change, normal julia promotion rules apply, and values will only be promoted as needed.

First adding tests to confirm that this behavior is currently broken, then will push teh fix and open for review when CI is green.

### DIFF
--- a/test/modelmatrix.jl
+++ b/test/modelmatrix.jl
@@ -409,6 +409,10 @@
         y, x = modelcols(apply_schema(f, schema(d)), d)
         @test eltype(y) == eltype(x) == Float32
 
+        f0 = @formula(y ~ 0 + x + log(x) * z)
+        y, x = modelcols(apply_schema(f0, schema(d)), d)
+        @test eltype(y) == eltype(x) == Float32
+
         contr = DummyCoding(; base=4, levels=1:4)
         d = (; y=rand(Float32, 4), x=rand(Float32, 4), z=[1:4; ])
 
--- a/test/modelmatrix.jl
+++ b/test/modelmatrix.jl
@@ -402,5 +402,27 @@
         f = apply_schema(@formula(0 ~ a&b&c), schema(t))
         @test vec(modelcols(f.rhs, t)) == modelcols.(Ref(f.rhs), Tables.rowtable(t))
     end
+
+    @testset "#293 conversion of Float32s" begin
+        d = (; y=rand(Float32, 4), x=rand(Float32, 4), z=[1:4; ])
+        f = @formula(y ~ 1 + x + log(x) * z)
+        y, x = modelcols(apply_schema(f, schema(d)), d)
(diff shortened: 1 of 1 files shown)


palday:
I'm generally onboard with this, but we should probably do something like a pkgeval to see if we would actually break anything. But we're also not so near the bottom of the dependency tree as StatsBase so a formally breaking release wouldn't be as bad as watching e.g. the StatsBase 0.34 changes propagate.

test/modelmatrix.jl
```diff
@@ -402,5 +402,35 @@
...
+        f0 = @formula(y ~ 0 + x + log(x) * z)
+        y, x = modelcols(apply_schema(f0, schema(d)), d)
+        @test eltype(y) == eltype(x) == Float32
+
+        fint = @formula(y ~ 1 + z)
+        y, x = modelcols(apply_schema(fint, schema(d)), d)
+        @test eltype(x) == Int
+
+        contr = DummyCoding(; base=4, levels=1:4)
+        d = (; y=rand(Float32, 4), x=rand(Float32, 4), z=[1:4; ])
```
palday, line 421: this isn't yet used below, right?

kleinschmidt: it is used but it's redundant with the definition above...

kleinschmidt: oh wait the contr part?  yeah that's right.

nalimilan:
+1

We could tag 1.0, as it was the plan already (removing `TableRegressionModel`): https://github.com/JuliaStats/StatsModels.jl/issues/271.

src/terms.jl
```diff
@@ -547,8 +547,10 @@ function modelcols(t::InteractionTerm, d::ColumnTable)
     row_kron_insideout(*, (modelcols(term, d) for term in t.terms)...)
 end
 
-modelcols(t::InterceptTerm{true}, d::NamedTuple) = ones(size(first(d)))
-modelcols(t::InterceptTerm{false}, d) = Matrix{Float64}(undef, size(first(d),1), 0)
+# use Bool as the eltype here to avoid promoting e.g. Float32s in other columns
+# to Float64 unless it's really necessary
+modelcols(t::InterceptTerm{true}, d::NamedTuple) = ones(Bool, size(first(d)))
+modelcols(t::InterceptTerm{false}, d) = Matrix{Bool}(undef, size(first(d),1), 0)
```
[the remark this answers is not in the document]
kleinschmidt: I'd kinda hoped to be able to just use a scalar but hcat doesn't work with that

andreasnoack:
Is there anything holding this one back at this point?

PR #299, opened 2023-09-02
termnames

palday:
xref https://github.com/JuliaStats/GLM.jl/issues/428


src/statsmodel.jl
```diff
@@ -106,12 +106,50 @@ function formula end
...
+    termnames(model::StatisticalModel)
+    termnames(term::AbstractTerm)
+
+Return the names associated with terms associated with a model.
+
+For models with only continuous predictors, this is the same as
+`(responsename(model), coefnames(model))`.
+
+For models with categorical predictors, the returned names reflect
+the categorical predictor and not the coefficients resulting from
```
nalimilan, line 119:
```suggestion
the variable name and not the coefficients resulting from
```

src/statsmodel.jl
```diff
@@ -106,12 +106,50 @@ function formula end
 formula(m::TableStatisticalModel) = m.mf.f
 formula(m::TableRegressionModel) = m.mf.f
 
+"""
+    termnames(model::StatisticalModel)
+    termnames(term::AbstractTerm)
+
+Return the names associated with terms associated with a model.
```
nalimilan, line 113:
```suggestion
Return the names of terms used in the formula of `model`.
```
It would be useful to specify whether this always returns a tuple, or whether for models with no response (e.g. PCA) this will return a vector without wrapping it in a tuple.

src/statsmodel.jl
```diff
@@ -106,12 +106,50 @@ function formula end
 formula(m::TableStatisticalModel) = m.mf.f
 formula(m::TableRegressionModel) = m.mf.f
 
+"""
+    termnames(model::StatisticalModel)
+    termnames(term::AbstractTerm)
```
nalimilan, line 111: This is the next docstring.
```suggestion
```

src/statsmodel.jl
```diff
@@ -106,12 +106,50 @@ function formula end
...
+
+Return the names associated with terms associated with a model.
+
+For models with only continuous predictors, this is the same as
```
nalimilan, line 115: Only regression models have `responsename`:
```suggestion
For regression models with only continuous predictors, this is the same as
```

src/statsmodel.jl
```diff
@@ -106,12 +106,50 @@ function formula end
...
+`String` or an iterable of `String`s.
+"""
+termnames(t::FormulaTerm) = (termnames(t.lhs), termnames(t.rhs))
+termnames(::InterceptTerm{H}) where {H} = H ? "(Intercept)" : nothing
+termnames(t::ContinuousTerm) = string(t.sym)
+termnames(t::CategoricalTerm) = string(t.sym)
+termnames(t::Term) = string(t.sym)
+termnames(t::ConstantTerm) = string(t.n)
+termnames(t::FunctionTerm) = string(t.exorig)
+termnames(ts::TupleTerm) = reduce(vcat, termnames.(ts))
```
nalimilan, line 142:
```suggestion
termnames(ts::TupleTerm) = mapreduce(termnames, vcat, ts)
```

src/statsmodel.jl
```diff
@@ -106,12 +106,50 @@ function formula end
...
+termnames(t::InteractionTerm) =
+    kron_insideout((args...) -> join(args, " & "), vectorize.(termnames.(t.terms))...)
+
 @doc """
     fit(Mod::Type{<:StatisticalModel}, f::FormulaTerm, data, args...;
         contrasts::Dict{Symbol}, kwargs...)
 
 Convert tabular data into a numeric response vector and predictor matrix using
-the formula `f`, and then `fit` the specified model type, wrapping the result in
+the formula `f`, and then `fit` the specified model type, wrapping Stthe result in
```
nalimilan, line 152:
```suggestion
the formula `f`, and then `fit` the specified model type, wrapping the result in
```

test/statsmodel.jl
```diff
@@ -233,6 +235,13 @@ Base.show(io::IO, m::DummyModTwo) = println(io, m.msg)
     m2 = fit(DummyModTwo, f, d)
     # make sure show() still works when there is no coeftable method
     show(io, m2)
+
+    # one final termnames check
+    # note that `1` is still a ConstantTerm and not yet InterceptTerm
+    # because apply_schema hasn't been called
```
nalimilan, line 241: Maybe this is worth mentioning in the docstring?

test/statsmodel.jl
```diff
@@ -233,6 +235,13 @@ Base.show(io::IO, m::DummyModTwo) = println(io, m.msg)
...
+    @test termnames(@formula(y ~ 1 + log(x) * y + (1+x|g)))[2] == 
+          ["1", "log(x)", "y", "log(x) & y", "(1 + x) | g"]
+          
 end
```
nalimilan, line 245: Can you add tests that call `termnames` directly on individual terms?

test/statsmodel.jl
palday: done

test/statsmodel.jl
palday: done

src/statsmodel.jl
```diff
@@ -106,6 +106,71 @@ function formula end
...
+
+For `RegressionModel`s with only continuous predictors, this is the same as
+`(responsename(model), coefnames(model))`.
+
+For models with categorical predictors, the returned names reflect
+the variable name and not the coefficients resulting from
+the choice of contrast coding.
+"""
+termnames(model::StatisticalModel) = termnames(formula(model))
+"""
```
nalimilan, line 124:
```suggestion

"""
```

src/statsmodel.jl
```diff
@@ -106,6 +106,71 @@ function formula end
...
+julia> termnames(@formula(y ~ x * y + (1+x|g)))
+("y", ["x", "y", "x & y", "(1 + x) | g"])
+```
+"""
+termnames(t::FormulaTerm) = (termnames(t.lhs), termnames(t.rhs))
+
+"""
+    termnames(term::AbstractTerm)
+
+Return the name(s) of column(s) generated by a term.
```
nalimilan, line 153: This isn't true for categorical terms, is it?

test/statsmodel.jl
```diff
@@ -232,7 +234,30 @@ Base.show(io::IO, m::DummyModTwo) = println(io, m.msg)
...
+    cm = StatsModels.ContrastsMatrix([1 0; 0 1], ["b", "c"], ["a", "b", "c"], DummyCoding())
+    @test termnames(CategoricalTerm(:x, cm)) =="x"
+    @test termnames(FunctionTerm(log, [Term(:x)], :(log(x)))) == "log(x)"
+    # these next few seem a little weird but they're consistent with the 
+    # definition of coefnames
+    @test termnames(InteractionTerm(term.((:a, :b, :c)))) == ["a & b & c"]
+    @test termnames(MatrixTerm(term(:a))) == "a"
+    @test termnames(MatrixTerm((term(:a), term(:b)))) == ["a", "b"]
+    @test termnames((term(:a), term(:b))) == ["a", "b"]
+    @test termnames((term(:a),)) == "a"
```
nalimilan, line 260: @kleinschmidt Is there any particular reason why `coefnames` behaves that way? If not, maybe we would better adopt a saner behavior for `termnames`, in the perspective of changing that of `coefnames` later?

src/statsmodel.jl
```diff
@@ -106,6 +106,71 @@ function formula end
...
+For models with categorical predictors, the returned names reflect
+the variable name and not the coefficients resulting from
+the choice of contrast coding.
+"""
+termnames(model::StatisticalModel) = termnames(formula(model))
+"""
+    termnames(t::FormulaTerm)
+
+Return a two-tuple of `termnames` applied to the left and
+right hand sides of the formula.
```
nalimilan, line 128: Given that we guarantee that this returns a two-tuple and that the method above is a shorthand for `termnames(formula(model))`, shouldn't we also say in the docstring for that method that it returns a two-tuple? (I hadn't realized, but at least currently formulas are required to have both a LHS and a RHS.)

src/contrasts.jl
```diff
@@ -87,7 +87,7 @@ mutable struct MyCoding <: AbstractContrasts
 end
 
 contrasts_matrix(C::MyCoding, baseind, n) = ...
-termnames(C::MyCoding, levels, baseind) = ...
+_termnames(C::MyCoding, levels, baseind) = ...
```
nalimilan, line 90: Isn't this breaking given that it's a documented function? It's kind of embarrassing given that it returns the names of generated columns/coefficients, not those of terms. Maybe we could rename it to something like `colnames` or `coefnames`, with a deprecation?

palday: Ugh, yeah I missed that it's in the [manual](https://juliastats.org/StatsModels.jl/stable/contrasts/#StatsModels.AbstractContrasts). Yes we should deprecate and make it a method of `coefnames`.

src/statsmodel.jl
palday: copy-paste mistake -- this is very lightly adapted from the code for `coefnames`.

test/statsmodel.jl
palday: I think it's an oversight in `coefnames` that rarely matters -- it can be 'fixed' by passing `init=[]` to `mapreduce`. The one spot where there may be a real difference is `InteractionTerm` because that may expand to multiple coefs/coefnames, but it's still one term/termname.

src/contrasts.jl
```diff
@@ -103,7 +103,7 @@ abstract type AbstractContrasts end
 # Contrasts + Levels (usually from data) = ContrastsMatrix
 struct ContrastsMatrix{C <: AbstractContrasts, M <: AbstractMatrix, T, U}
     matrix::M
-    termnames::Vector{U}
+    termnames::Vector{U} # XXX this is somewhat of a misnomer, this should be coefnames...
```
nalimilan, line 106: You could rename the field but define a deprecated `getproperty` method to keep allowing `obj.termnames`.

src/contrasts.jl
```diff
@@ -224,7 +224,11 @@ function ContrastsMatrix(c::ContrastsMatrix, levels::AbstractVector)
     return c
 end
 
-function _termnames(C::AbstractContrasts, levels::AbstractVector, baseind::Integer)
+@deprecate(termnames(C::AbstractContrasts, levels::AbstractVector, baseind::Integer),
+           coefnames(C::AbstractContrasts, levels::AbstractVector, baseind::Integer),
+           false)
```
nalimilan, line 229: Put this in a src/deprecated.jl file?

src/statsmodel.jl
```diff
@@ -75,7 +75,7 @@ end
 for (modeltype, dfmodeltype) in ((:StatisticalModel, TableStatisticalModel),
                                  (:RegressionModel, TableRegressionModel))
     @eval begin
-        function StatsBase.fit(::Type{T}, f::FormulaTerm, data, args...;
+        function StatsAPI.fit(::Type{T}, f::FormulaTerm, data, args...;
                                contrasts::Dict{Symbol,<:Any} = Dict{Symbol,Any}(),
                                kwargs...) where T<:$modeltype
```
nalimilan, line 80:
```suggestion
        function StatsAPI.fit(::Type{T}, f::FormulaTerm, data, args...;
                              contrasts::Dict{Symbol,<:Any} = Dict{Symbol,Any}(),
                              kwargs...) where T<:$modeltype
```

src/statsmodel.jl
```diff
@@ -121,14 +122,16 @@ the variable name and not the coefficients resulting from
...
 Return a two-tuple of `termnames` applied to the left and
 right hand sides of the formula.
- 
-Note that until `apply_schema` has been called, literal `1` in formulae
-is interpreted as `ConstantTerm(1)` and will thus appear as `"1"` in the 
+
+!!! note
+    Until `apply_schema` has been called, literal `1` in formulae
+is interpreted as `ConstantTerm(1)` and will thus appear as `"1"` in the
 returned term names.
```
nalimilan:
```suggestion
    Until `apply_schema` has been called, literal `1` in formulae
    is interpreted as `ConstantTerm(1)` and will thus appear as `"1"` in the
    returned term names.
```

src/statsmodel.jl
```diff
@@ -161,15 +164,15 @@ termnames(t::CategoricalTerm) = string(t.sym)
...
-termnames(ts::TupleTerm) = mapreduce(termnames, vcat, ts)
-# these have some surprising behavior:
-# termnames(::InteractionTerm) always returns a vector
-# termnames(MatrixTerm(term(:a))) returns a scalar
-# termnames(MatrixTerm((term(a:), term(:b)))) returns a vector
-# but this is the same behavior as coefnames
-termnames(t::MatrixTerm) = mapreduce(termnames, vcat, t.terms)
-termnames(t::InteractionTerm) =
-    kron_insideout((args...) -> join(args, " & "), vectorize.(termnames.(t.terms))...)
+# termnames(TupleTerm)) alwyas returns a vector, even if it's just one element, e.g.,
```
nalimilan, line 167:
```suggestion
# termnames(TupleTerm)) always returns a vector, even if it's just one element, e.g.,
```

src/statsmodel.jl
```diff
@@ -161,15 +164,15 @@ termnames(t::CategoricalTerm) = string(t.sym)
...
+# termnames((term(:a),))
+termnames(ts::TupleTerm) = mapreduce(termnames, vcat, ts; init=[])
+# termnames(MatrixTerm)) alwyas returns a vector, even if it's just one element, e.g.,
```
nalimilan, line 170:
```suggestion
# termnames(MatrixTerm)) always returns a vector, even if it's just one element, e.g.,
```

test/statsmodel.jl
```diff
@@ -8,17 +8,17 @@ struct DummyMod <: RegressionModel
...
-StatsBase.coeftable(mod::DummyMod) =
+StatsAPI.coeftable(mod::DummyMod) =
     CoefTable(reshape(mod.beta, (size(mod.beta,1), 1)),
               ["'beta' value"],
               ["" for n in 1:size(mod.x,2)],
               0)
 # dumb predict: return values predicted by "beta" and dummy confidence bounds
-function StatsBase.predict(mod::DummyMod;
+function StatsAPI.predict(mod::DummyMod;
                            interval::Union{Nothing,Symbol}=nothing)
```
nalimilan, line 22:
```suggestion
                          interval::Union{Nothing,Symbol}=nothing)
```

test/statsmodel.jl
```diff
@@ -29,7 +29,7 @@ function StatsBase.predict(mod::DummyMod;
         throw(ArgumentError("value not allowed for interval"))
     end
 end
-function StatsBase.predict(mod::DummyMod, newX::Matrix;
+function StatsAPI.predict(mod::DummyMod, newX::Matrix;
                            interval::Union{Nothing,Symbol}=nothing)
```
nalimilan, line 33:
```suggestion
                          interval::Union{Nothing,Symbol}=nothing)
```

src/statsmodel.jl
```diff
@@ -106,6 +106,74 @@ function formula end
...
+termnames(t::FunctionTerm) = string(t.exorig)
+# termnames(TupleTerm)) alwyas returns a vector, even if it's just one element, e.g.,
+# termnames((term(:a),))
+termnames(ts::TupleTerm) = mapreduce(termnames, vcat, ts; init=[])
+# termnames(MatrixTerm)) alwyas returns a vector, even if it's just one element, e.g.,
+# termnames(MatrixTerm(term(:a)))
+termnames(t::MatrixTerm) = mapreduce(termnames, vcat, t.terms; init=[])
+function termnames(t::InteractionTerm)
+    only(kron_insideout((args...) -> join(args, " & "), vectorize.(termnames.(t.terms))...))
+end
```
nalimilan, line 175:
```suggestion
termnames(t::InteractionTerm) =
    only(kron_insideout((args...) -> join(args, " & "), vectorize.(termnames.(t.terms))...))
```

src/contrasts.jl
palday: done

src/contrasts.jl
palday: done

src/contrasts.jl
```diff
@@ -224,7 +224,8 @@ function ContrastsMatrix(c::ContrastsMatrix, levels::AbstractVector)
     return c
 end
 
-function termnames(C::AbstractContrasts, levels::AbstractVector, baseind::Integer)
+
```
nalimilan, line 227:
```suggestion
```

nalimilan (approved):

palday:
@kleinschmidt GH won't let me reply to the comment so....

> I'm a bit worried by teh inconsistency with `coefnames` here, where an empty vector is used. Also, `vcat` with `nothing` will add an explicit `nothing` into the result. but perhaps that's intentional?

Yeah, I don't think that most users should call these methods directly, and `String[]` plays nicer with `vcat`, which is called if applying this to a collection of terms / formula. I'll update when we get Milan's thoughts on the other bit of bikeshed. 😄

nalimilan:
`variablenames` could mean something different: in R, `all.vars(f)` returns the names of variables which are used in a formula, which allows doing e.g. subset(df, all.vars(f))` to keep only useful variables. `termnames` sounds closer to what we want but I would have to double-check the formula machinery to have a definitive opinion. `MatrixTerm` and `TupleTerm` are kind of meta-terms wrapping other terms, right? So it makes sense that they don't have names themselves.

src/statsmodel.jl
```diff
@@ -106,6 +106,73 @@ function formula end
...
+This is a convenience method for `termnames(formula(model))`, which returns a
+two-tuple of `termnames` applied to the left and right hand sides of the formula.
+
+For `RegressionModel`s with only continuous predictors, this is the same as
+`(responsename(model), coefnames(model))`.
+
+For models with categorical predictors, the returned names reflect
+the variable name and not the coefficients resulting from
+the choice of contrast coding.
+"""
```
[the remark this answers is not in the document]
palday: done

src/statsmodel.jl
```diff
@@ -106,6 +106,73 @@ function formula end
...
+termnames(t::ConstantTerm) = string(t.n)
+termnames(t::FunctionTerm) = string(t.exorig)
+# termnames(TupleTerm)) always returns a vector, even if it's just one element, e.g.,
+# termnames((term(:a),))
+termnames(ts::TupleTerm) = mapreduce(termnames, vcat, ts; init=String[])
+# termnames(MatrixTerm)) always returns a vector, even if it's just one element, e.g.,
+# termnames(MatrixTerm(term(:a)))
+termnames(t::MatrixTerm) = mapreduce(termnames, vcat, t.terms; init=String[])
+termnames(t::InteractionTerm) =
+    only(kron_insideout((args...) -> join(args, " & "), vectorize.(termnames.(t.terms))...))
```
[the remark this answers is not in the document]
palday: done

palday:
I'm open to other names -- it's easy enough to change `variablenames` to anything else. The hard part was `termnames` to `variablenames` because that wasn't a universal replacement.

src/deprecated.jl
```diff
@@ -0,0 +1,12 @@
...
+           coefnames(C::AbstractContrasts, levels::AbstractVector, baseind::Integer),
+           false)
+
+function Base.getproperty(cm::ContrastsMatrix, x::Symbol)
+    if x === :termnames 
+        Base.depwarn("The `termnames` field has been renamed `coefnames`.", :ContrastsMatrix)
+        x = :coefnames
+    end
+
+    return getfield(cm, x)
```
ararslan, line 11:
```suggestion
    if x === :termnames
        Base.depwarn("the `termnames` field of `ConstrastsMatrix` is deprecated; use `coefnames(cm)` instead."
        return coefnames(cm)
    else
        return getfield(cm, x)
    end
```
Might as well make the deprecation target encourage better behavior

palday:
@kleinschmidt @nalimilan we could also go with the hybrid function name `termvarnames`. Or make this a completely internal method `_variablenames` because we're really only using it for `gvif`. We can always make it public later.

palday:
okay, it seems like @kleinschmidt is now onboard with termnames. Great!

kleinschmidt (approved):
looks like docs are not picking up the termnames crossref: https://github.com/JuliaStats/StatsModels.jl/actions/runs/6102408363/job/16561033364?pr=299#step:5:279

fix that and then LGTM.

nalimilan:
Looks like you haven't noticed my last comment at https://github.com/JuliaStats/StatsModels.jl/pull/299#discussion_r1317837975. Not a big deal, but maybe still worth adjusting?

src/statsmodel.jl
```diff
@@ -106,6 +106,73 @@ function formula end
...
+
+"""
+    termnames(term::AbstractTerm)
+
+Return the name of a term.
+
+Return value is either a `String`, an iterable of `String`s or nothing if there
+no associated name (e.g. `termnames(InterceptTerm{false}())`).
+"""
+termnames(::InterceptTerm{H}) where {H} = H ? "(Intercept)" : nothing
```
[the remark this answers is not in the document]
palday: @nalimilan we still need it to return something so that we get don't get an error for `y ~ 0 + x` where there is a valid term names for the whole entity. The problem with nothing comes in the map-reduce step for compound terms.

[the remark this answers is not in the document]
palday: I guess one other option would be to have `termnames` *always* return a String array, which could be empty (e.g. here), contain one element (ContinuousTerm) or multiple elements (CategoricalTerm). We haven't tagged the release yet, so now's the time to do the change :wink:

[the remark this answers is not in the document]
nalimilan: I hadn't realized you had changed it to return `String[]` instead of  `nothing`. There's no inconsistency with `coefnames` anymore then. Let's keep it that way.

PR #300, opened 2023-09-06
VIF and GVIF

palday:
- [x] still needs more extensive tests, but I haven't had a chance to see what it would take to fake something reasonable


src/vif.jl
```diff
@@ -0,0 +1,119 @@
...
+    x = collect_matrix_terms(x)
+    x = x isa MatrixTerm ? x : first(x)
+    x isa MatrixTerm || throw(ArgumentError("couldn't extract matrix term from $x"))
+    if first(x.terms) isa MatrixTerm
+        x = only(x.terms)
+    end
+    return x
+end
+
+"""$(@doc StatsAPI.vif)"""
```
nalimilan, line 31: If we repeat the docstring it will be shown twice, right?

test/runtests.jl
```diff
@@ -23,11 +23,15 @@ my_tests = ["ambiguity.jl",
...
+            "protect.jl",
+            "vif.jl"]
 
 @testset "StatsModels" begin
     @testset "aqua" begin
-        Aqua.test_all(StatsModels; ambiguities=false)
+        # because VIF and GVIF are defined in StatsAPI for RegressionModel,
+        # which is also defined there, it's flagged as piracy. But
+        # we're the offical implementers so it's privateering.
+        Aqua.test_all(StatsModels; ambiguities=false, piracy=false)
```
nalimilan, line 34: Looks like you can pass something like `ambiguities=(exclude=[vif, gvif],)`.

src/vif.jl
```diff
@@ -0,0 +1,119 @@
+# putting this behind a function barrier means that the model matrix
+# can potentially be freed immediately if it's large and constructed on the fly
+function _find_intercept(model::RegressionModel)
+    modelmat = modelmatrix(model)
+    # XXX collect is necessary for Julia 1.6
+    # but it's :just: an array of references to views, so shouldn't be too
+    # expensive
+    return findfirst(Base.Fix1(all, isone), collect(eachcol(modelmat)))
```
nalimilan, line 8: How about using a `VERSION` check? That way we won't forget dropping it when we drop Julia 1.6 support and we don't waste memory on newer Julia versions.

src/vif.jl
palday: This is the only way I found to get it to show up in the manual here.

nalimilan: [Apparently](https://stackoverflow.com/questions/70137119/how-to-include-the-docstring-for-a-function-from-another-package-in-my-julia-doc) you can pass `modules=[StatsModels, StatsAPI]` to `makedocs`. This will probably have the side effect of printing warnings for StatsAPI docstrings that are not in the manual though, but that's less of an issue IMO.

src/vif.jl
palday: done

src/vif.jl
palday: oh very nice, done

test/runtests.jl
palday: That works for `ambiguities` but not `piracy` :frowning:

src/vif.jl
```diff
@@ -44,10 +47,11 @@ function StatsAPI.vif(model::RegressionModel)
         throw(ArgumentError("VIF not meaningful for models with only one non-intercept term"))
     # NB: The correlation matrix is positive definite and hence invertible
     #     unless there is perfect rank deficiency, hence the warning in the docstring.
-    return diag(inv(m))
+    # there's no `inv(::SubArray)` method before 1.7
+    # but this matrix is indeed symmetric and the Symmetric wrapper is enough to make this work
+    return diag(inv(Symmetric(m)))
```
nalimilan, line 52: Is there any performance advantage in using `Symmetric`? Otherwise I'd also use a `VERSION` check to avoid using more complex code on recent Julia versions.

src/vif.jl
```diff
@@ -93,10 +97,11 @@ function StatsAPI.gvif(model::RegressionModel; scale=false)
...
         throw(ArgumentError("GVIF not meaningful for models with only one non-intercept term"))
 
-    tn = termnames(model)
+    tn = last(termnames(model))
     tn = view(tn, axes(tn, 1) .!= intercept)
-    trms = get_matrix_term(form)
-    trms = view(trms, axes(trms, 1) .!= intercept)
+    trms = get_matrix_term(form.rhs).terms
+    # MatrixTerms.terms is a tuple or vector so always 1-based indexing
+    trms = [trms[i] for i in 1:length(trms) if i != intercept]
```
nalimilan, line 104: Simpler and usually faster:
```suggestion
    trms = trms[1:length(trms) .!= intercept]
```

test/vif.jl
```diff
@@ -18,7 +18,23 @@ end
...
+    modelmatrix::Matrix{Float64}
+    vcov::Matrix{Float64}
+    formula::FormulaTerm
+end
+
+for f in [:coefnames, :coef, :stderror, :modelmatrix, :vcov]
+    @eval StatsAPI.$f(model::Duncan) = model.$f
+end 
+
+# StatsModels.termnames(model::Duncan) = model.termnames
```
nalimilan, line 34: ?

test/vif.jl
```diff
@@ -28,3 +44,38 @@ StatsAPI.vcov(::MyRegressionModel3) = [1 0 0; 0 1 0; 0 0 1]
...
+                           -0.151845   0.0143203   -0.00851855
+                           -0.150706  -0.00851855   0.00965358])
+    @test vif(duncan2) ≈ [2.1049, 2.1049] atol=1e-5
+    # two different ways of calculating the same quantity
+    @test vif(duncan2) ≈ gvif(duncan2) 
+end
+
+@testset "GVIF" begin
+    cm = StatsModels.ContrastsMatrix(DummyCoding("bc", ["bc", "prof", "wc"]), ["bc", "prof", "wc"])
+    duncan3 =  Duncan(; coefnames=["(Intercept)", "Income", "Education",  "Type: prof", "Type: wc"],
```
nalimilan, line 67:
```suggestion
    duncan3 = Duncan(; coefnames=["(Intercept)", "Income", "Education",  "Type: prof", "Type: wc"],
```

test/vif.jl
```diff
@@ -28,3 +44,38 @@ StatsAPI.vcov(::MyRegressionModel3) = [1 0 0; 0 1 0; 0 0 1]
...
+                     stderror=[3.71377, 0.0893553, 0.113609, 6.99301, 6.10877],
+                     # we actually don't need the whole matrix -- just enough to find an intercept
+                     modelmatrix=[1.0 62.0 86.0 1.0 0.0], 
+                     vcov=[13.7921    -0.115637    -0.257486    14.0947     7.9022
+                           -0.115637   0.00798437  -0.00292449  -0.126011  -0.109049
+                           -0.257486  -0.00292449   0.012907    -0.616651  -0.38812
+                           14.0947    -0.126011    -0.616651    48.9021    30.2139
+                            7.9022    -0.109049    -0.38812     30.2139    37.3171])
+    @test gvif(duncan3) ≈ [2.209178, 5.297584, 5.098592] atol=1e-4
+    @test gvif(duncan3; scale=true) ≈ [1.486330, 2.301648, 1.502666] atol=1e-5
```
nalimilan, line 80: I assume you checked these against another implementation? :-)

test/runtests.jl
nalimilan: What about `piracy=(treat_as_own=[vif, gvif],)` or `piracy=(treat_as_own=[RegressionModel],)`? The Aqua manual seems tailored to this PR: :-D
> treat_as_own = Union{Function, Type}[]: The types in this container are considered to be "owned" by the module m. This is useful for testing packages that deliberately commit some type piracy, e.g. modules adding higher-level functionality to a lightweight C-wrapper, or packages that are extending StatsAPI.jl.

nalimilan:
Also there seem to remain a few uncovered lines (apart from the `@static if VERSION` which is obviously a bug).

test/vif.jl
palday:
```suggestion
```

test/vif.jl
palday: Yep, against `car` back when I originally wrote these tests for MixedModelsExtras. :smile:

src/vif.jl
palday: There should be -- the inverse of a symmetric matrix is itself symmetric, so you can take advantage of that to get better performance. (In other words, if LinearAlgebra doesn't take advantage of this, it should and may in the future.)

test/runtests.jl
palday: merci! updated

src/vif.jl
nalimilan: OK. In that case let's just drop the comment which seems to imply we only do this as a workaround:
```suggestion
    return diag(inv(Symmetric(m)))
```

src/vif.jl
```diff
@@ -26,6 +26,7 @@ function get_matrix_term(x)
     x = collect_matrix_terms(x)
     x = x isa MatrixTerm ? x : first(x)
     x isa MatrixTerm || throw(ArgumentError("couldn't extract matrix term from $x"))
+    # this is a work around for some weirdness that happens in MixedModels.jl
```
nalimilan, line 29: Is there a way to test this here? It doesn't seem covered.

palday: tested by manually constructing such a MatrixTerm.

I don't want to take MixedModels (or GLM for that matter) as a test dependency because then we'll have trouble testing breaking releases because we won't be able to install the reverse dependencies.

nalimilan (approved):

test/vif.jl
kleinschmidt: would be good to have _some_ back link in case we need to check/update these (same for the testset above)

kleinschmidt (approved):
I'd generally be a lot more comfortable if there was some kind of reproducible way to get these reference values (even if it's in a script that doesn't actually get automatically executed during tests).  but otherwise looks good

Project.toml
```diff
@@ -16,13 +16,13 @@ StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 
 [compat]
-Aqua = "0.6"
+Aqua = "0.6, 0.7"
```
palday, line 19:
```suggestion
Aqua = "0.7"
```

test/vif.jl
```diff
@@ -0,0 +1,91 @@
...
+    # no intercept term
+    @test_throws ArgumentError vif(MyRegressionModel())
+
+    # only one non intercept term
+    @test_throws ArgumentError vif(MyRegressionModel2())
+
+    # vcov is identity, so the VIF is just 1
+    @test vif(MyRegressionModel3()) ≈ [1, 1]
+end
+
```
palday, line 46:
```suggestion

# Reference values from
# https://github.com/palday/MixedModelsExtras.jl/blob/6db061cb856301c04b978101d4c53b90c890715b/test/vif.jl
# which were computed using car::vif in R

```

test/vif.jl
```diff
@@ -44,6 +44,10 @@ StatsModels.formula(model::Duncan) = model.formula
     @test vif(MyRegressionModel3()) ≈ [1, 1]
 end
 
+# Reference values from
+# https://github.com/palday/MixedModelsExtras.jl/blob/6db061cb856301c04b978101d4c53b90c890715b/test/vif.jl
```
nalimilan, line 48: Can you copy the relevant parts from that file instead? It's weird to rely on MixedModelsExtras for that, especially if this PR is supposed to provide the same features more upstream.

palday: I mean, the relevant parts of that file are just the reference values. I'm planning on writing out the R code to do it again, which is why I hadn't merged this PR yet.

palday: added in https://github.com/JuliaStats/StatsModels.jl/pull/300/commits/96686ab44cdb921e14a120be8ea90ecb544237c4

PR #301, opened 2023-09-07
Slighltly improve `coefnames` with no intercept

nalimilan:
Adopt improvements from `termnames` (https://github.com/JuliaStats/StatsModels.jl/pull/299).


palday (approved):

src/terms.jl
```diff
@@ -567,13 +567,15 @@ vectorize(x) = [x]
...
+Return the name(s) of column(s) generated by a term.
+
+Return value is either a `String`, an iterable of `String`s or the empty vector 
+if there is no associated column (e.g. `coefnames(InterceptTerm{false}())`).
 
 See also [`termnames`](@ref).
 """
 StatsAPI.coefnames(t::FormulaTerm) = (coefnames(t.lhs), coefnames(t.rhs))
-StatsAPI.coefnames(::InterceptTerm{H}) where {H} = H ? "(Intercept)" : []
+StatsAPI.coefnames(::InterceptTerm{H}) where {H} = H ? "(Intercept)" : String[]
```
ararslan, line 578: Tangential as this code is predates this PR but it's kind of odd IMO that `coefnames` can return a `String` rather than a container. The name is plural so it sounds like it should give you some collection of names, even if said collection has one element (or none).

palday: I would not be opposed to always returning a container, even when it's just one element. (This code hasn't been released yet, so it's not yet breaking.)

nalimilan: It's true that the name is plural, but I'm not sure returning a single string is a problem in practice. Are people really likely to call this on objects which could either be single- or multiple-term? It's not super practical either to get a single-element vector when you know in advance there will be a single value, and in terms of our own implementation it's quite wasteful (though it probably doesn't matter much).

kleinschmidt: teh return type is already all over the place since the `FormulaTerm` one returns a tuple, so I'm not too fussed about it.

kleinschmidt: I guess that means the current docstring is wrong though...

src/terms.jl
```diff
@@ -567,13 +567,15 @@ vectorize(x) = [x]
 """
     coefnames(term::AbstractTerm)
 
-Return the name(s) of column(s) generated by a term.  Return value is either a
-`String` or an iterable of `String`s.
+Return the name(s) of column(s) generated by a term.
+
+Return value is either a `String`, an iterable of `String`s or the empty vector 
+if there is no associated column (e.g. `coefnames(InterceptTerm{false}())`).
```
kleinschmidt, line 573:
```suggestion
Return value is either a `String`, an iterable of `String`s, the empty vector 
if there is no associated column (e.g. `coefnames(InterceptTerm{false}())`), or
(in the case of a `FormulaTerm`) a two tuple of any of the above (corresponding 
to `coefnames` applied to the left- and right-hand sides of the formula).
```

kleinschmidt (approved):
I'm fine with this overall, just a minor tweak to be super pedantic about the possible return values.

src/terms.jl
nalimilan: What if I copied and adapted the docstring for `termnames(t::FormulaTerm)` instead? My goal was to make the two functions consistent, and the behavior for formulas seems different enough to deserve a separate docstring.

src/terms.jl
matthieugomez: I agree with @ararslan — always returning a container would be the best choice (and that’s what Julia does everywhere)And yes, it matters for developers that use coefnames inside their functions

src/terms.jl
nalimilan: Bump.

PR #320, opened 2024-12-06
Fix issue with \times in docstring

rasmushenningsson:
Replaced \times with × that will work in both docstring and docs.

### DIFF
--- a/src/contrasts.jl
+++ b/src/contrasts.jl
@@ -471,7 +471,7 @@ end
     HypothesisCoding(hypotheses::AbstractMatrix; levels=nothing, labels=nothing)
 
 Specify how to code a categorical variable in terms of a *hypothesis matrix*.
-For a variable with ``k`` levels, this should be a ``k-1 \times k`` matrix.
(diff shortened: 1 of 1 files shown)


src/contrasts.jl
```diff
@@ -471,7 +471,7 @@ end
     HypothesisCoding(hypotheses::AbstractMatrix; levels=nothing, labels=nothing)
 
 Specify how to code a categorical variable in terms of a *hypothesis matrix*.
-For a variable with ``k`` levels, this should be a ``k-1 \times k`` matrix.
+For a variable with ``k`` levels, this should be a ``k-1×k`` matrix.
```
ajinkya-k, line 474:
```suggestion
For a variable with ``k`` levels, this should be a ``(k-1) × k`` matrix.
```

ajinkya-k: could also use `(k-1) \\times k` (note the double backslash)

ajinkya-k (commented):
thanks for the PR, I just literally added this in a different PR and rolled it back haha

nalimilan:
Thanks for spotting and fixing this!

PR #324, opened 2025-08-01
Bump CategoricalArrays to 1

nalimilan:
Since `levels(::CategoricalArray)` now returns a `CategoricalArray`, we need to
unwrap the result before storing it as an `Array` field. This also works
on CategoricalArrays 0.10.


nalimilan:
Thanks! I've bumped the version and filed https://github.com/JuliaRegistries/General/pull/136067.

PR #333, opened 2025-12-20
Update to Aqua@0.8

andreasnoack:
Supersedes https://github.com/JuliaStats/StatsModels.jl/pull/331


test/modelmatrix.jl
```diff
@@ -1,5 +1,5 @@
 @testset "Model matrix" begin
-    
+
```
devmotion: Different line endings? Are these changes intentional?

